PR #27, opened 2023-11-16
tests: address pixel tests flakiness with unloaded logo image

rvykydal:
Resolves: INSTALLER-3792

In my local tests the flakes disappeared, but some tests timeouted on the condition.


KKoukiou:
Rebased for CI to get green.

KKoukiou (approved):

KKoukiou (dismissed):

KKoukiou (commented):
I see indeed some screenshots still with non-loaded image. And the relevant tests failing.

KKoukiou (approved):
It does not seems to fix the issue - as I saw a fail run with this, but the check is correct so let's get it in.

PR #47, opened 2023-11-20
Validate user account value

rvykydal:
The PR was originally open on anaconda repository and has some history and discussion there: https://github.com/rhinstaller/anaconda/pull/5305

### DIFF
--- a/src/components/users/Accounts.jsx
+++ b/src/components/users/Accounts.jsx
@@ -19,10 +19,13 @@ import cockpit from "cockpit";
 import React, { useState, useEffect } from "react";
 import * as python from "python.js";
 import encryptUserPw from "../../scripts/encrypt-user-pw.py";
+import { debounce } from "throttle-debounce";
 
 import {
     Form,
     FormGroup,
+    FormHelperText,
+    HelperText,
     TextInput,
     Title,
 } from "@patternfly/react-core";
@@ -62,6 +65,33 @@ export const accountsToDbusUsers = (accounts) => {
     }];
 };
 
+const reservedNames = [
+    "root",
+    "bin",
+    "daemon",
+    "adm",
+    "lp",
+    "sync",
+    "shutdown",
+    "halt",
+    "mail",
+    "operator",
+    "games",
+    "ftp",
+    "nobody",
+    "home",
+    "system",
+];
+
+const isUserAccountWithInvalidCharacters = (userAccount) => {
+    return (
+        userAccount === "." ||
+        userAccount === ".." ||
+        userAccount.match(/^[0-9]+$/) ||
+        !userAccount.match(/^[A-Za-z0-9._][A-Za-z0-9._-]{0,30}([A-Za-z0-9._-]|\$)?$/)
+    );
+};
+
 const CreateAccount = ({
     idPrefix,
     passwordPolicy,
@@ -70,14 +100,39 @@ const CreateAccount = ({
     setAccounts,
 }) => {
     const [fullName, setFullName] = useState(accounts.fullName);
+    const [_userAccount, _setUserAccount] = useState(accounts.userAccount);
     const [userAccount, setUserAccount] = useState(accounts.userAccount);
+    const [userAccountInvalidHint, setUserAccountInvalidHint] = useState("");
+    const [isUserAccountValid, setIsUserAccountValid] = useState(null);
     const [password, setPassword] = useState(accounts.password);
     const [confirmPassword, setConfirmPassword] = useState(accounts.confirmPassword);
     const [isPasswordValid, setIsPasswordValid] = useState(false);
 
     useEffect(() => {
-        setIsUserValid(isPasswordValid && userAccount.length > 0);
-    }, [setIsUserValid, isPasswordValid, userAccount]);
+        debounce(300, () => setUserAccount(_userAccount))();
+    }, [_userAccount, setUserAccount]);
+
+    useEffect(() => {
+        setIsUserValid(isPasswordValid && isUserAccountValid);
+    }, [setIsUserValid, isPasswordValid, isUserAccountValid]);
+
+    useEffect(() => {
+        let valid = true;
+        setUserAccountInvalidHint("");
+        if (userAccount.length === 0) {
+            valid = null;
+        } else if (userAccount.length > 32) {
+            valid = false;
+            setUserAccountInvalidHint(_("User names must be shorter than 33 characters"));
+        } else if (reservedNames.includes(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(_("User name must not be a reserved word"));
+        } else if (isUserAccountWithInvalidCharacters(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(cockpit.format(_("User name may only contain: letters from a-z, digits 0-9, dash $0, period $1, underscore $2"), "-", ".", "_"));
+        }
+        setIsUserAccountValid(valid);
+    }, [userAccount]);
 
     const passphraseForm = (
         <PasswordFormFields
@@ -126,9 +181,15 @@ const CreateAccount = ({
             >
                 <TextInput
                   id={idPrefix + "-user-account"}
-                  value={userAccount}
-                  onChange={(_event, val) => setUserAccount(val)}
+                  value={_userAccount}
+                  onChange={(_event, val) => _setUserAccount(val)}
+                  validated={isUserAccountValid === null ? "default" : isUserAccountValid ? "success" : "error"}
                 />
+                <FormHelperText>
+                    <HelperText component="ul" aria-live="polite" id={idPrefix + "-full-name-helper"}>
+                        {userAccountInvalidHint}
+                    </HelperText>
+                </FormHelperText>
             </FormGroup>
             {passphraseForm}
         </Form>
--- a/src/components/users/Accounts.jsx
+++ b/src/components/users/Accounts.jsx
@@ -40,13 +40,13 @@ const _ = cockpit.gettext;
 
 export function getAccountsState (
     fullName = "",
-    userAccount = "",
+    userName = "",
     password = "",
     confirmPassword = "",
 ) {
     return {
         fullName,
-        userAccount,
+        userName,
         password,
         confirmPassword,
     };
@@ -59,7 +59,7 @@ export const cryptUserPassword = async (password) => {
 
 export const accountsToDbusUsers = (accounts) => {
     return [{
-        name: cockpit.variant("s", accounts.userAccount || ""),
+        name: cockpit.variant("s", accounts.userName || ""),
         gecos: cockpit.variant("s", accounts.fullName || ""),
         password: cockpit.variant("s", accounts.password || ""),
         "is-crypted": cockpit.variant("b", true),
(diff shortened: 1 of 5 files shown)


rvykydal:
@garret please see https://github.com/rhinstaller/anaconda/pull/5305#pullrequestreview-1737335987

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+    }, [setIsUserValid, isPasswordValid, isUserAccountValid]);
+
+    useEffect(() => {
+        let valid = true;
+        setUserAccountInvalidHint("");
+        if (userAccount.length === 0) {
+            valid = null;
+        } else if (userAccount.length > 32) {
+            valid = false;
+            setUserAccountInvalidHint(_("The user name is too long"));
```
garrett, line 119: Order of preference for the account name is:

1. Account name (doesn't use "user"; this is good, as "user" is bad to show in interfaces)
2. Username (the actual word, according to English dictionaries)
3. User name (an uncommon variant, according to English dictionaries)

As both GNOME and KDE talk about "user" accounts and have a "name" field, I guess keeping with "User name" is probably OK enough for consistency, even if it's not ideal.

Therefore, this is fine.

garrett: On the other hand, it would be good to mention what the cutoff is.

```suggestion
            setUserAccountInvalidHint(_("User names must contain fewer than 33 characters"));
```

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+        } else if (reservedNames.includes(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(_("Sorry, that user name is not available. Please try another."));
```
garrett, line 122: Since this is blocked exclusively because it is reserved word, it might be nice to say that specifically here.

```suggestion
            setUserAccountInvalidHint(_("This is a reserved word; please try another user name"));
```

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+        } else if (isUserAccountWithInvalidCharacters(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(cockpit.format(_("The user name should usually only consist of lower case letters from a-z, digits and the following characters: $0"), "-_"));
```
garrett, line 125: A few improvements:

- Rewording
- "lowercase" is one word
- Oxford comma for an inline list

```suggestion
            setUserAccountInvalidHint(cockpit.format(_("User names may only contain lowercase letters from a-z, digits, and the following characters: $0"), "-_"));
```

garrett (changes requested):
I've made a bunch of suggestions.

garrett:
⚠️ Also, in the future, when you're tagging me on GitHub, please use my name without typos. When you make typos, it makes other people upset... and I also won't see the ping.

(My name has two Rs and two Ts.)

rvykydal:
> ⚠️ Also, in the future, when you're tagging me on GitHub, please use my name without typos. When you make typos, it makes other people upset... and I also won't see the ping.
> 
> (My name has two Rs and two Ts.)

Sorry for that, relying on auto-completion makes me ignorant, now I'll remember :)

src/components/users/Accounts.jsx
rvykydal: Updated

src/components/users/Accounts.jsx
rvykydal: Updated

src/components/users/Accounts.jsx
rvykydal: Updated

rvykydal:
Screenshot from 2023-11-21 14-40-50
Screenshot from 2023-11-21 14-41-12
Screenshot from 2023-11-21 14-42-47

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+    }, [setIsUserValid, isPasswordValid, isUserAccountValid]);
+
+    useEffect(() => {
+        let valid = true;
+        setUserAccountInvalidHint("");
+        if (userAccount.length === 0) {
+            valid = null;
+        } else if (userAccount.length > 32) {
+            valid = false;
+            setUserAccountInvalidHint(_("User names must contain fewer than 33 characters"));
```
garrett, line 119:
```suggestion
            setUserAccountInvalidHint(_("User name must be shorter than 33 characters"));
```

garrett, line 119:
```suggestion
            setUserAccountInvalidHint(_("User name must be shorter than 33 characters"));
```

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+        } else if (reservedNames.includes(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(_("This is a reserved word; please try another user name"));
```
garrett, line 122:
```suggestion
            setUserAccountInvalidHint(_("User name must not be a reserved word"));
```

src/components/users/Accounts.jsx
```diff
@@ -63,14 +93,39 @@ const CreateAccount = ({
...
+        } else if (isUserAccountWithInvalidCharacters(userAccount)) {
+            valid = false;
+            setUserAccountInvalidHint(cockpit.format(_("User names may only contain lowercase letters from a-z, digits, and the following characters: $0"), "-._"));
```
garrett, line 125:
```suggestion
            setUserAccountInvalidHint(cockpit.format(_("User name may only contain: letters from a-z, digits 0-9, dash -, period ., and underscore _")));
```

garrett (changes requested):
Thanks for the screenshots!

I'm looking at this all again, in context now, with a clearer head (less of a post-COVID headache today), and I have another round of suggestions. Apologies!

This time around, I'll show high fidelity mockups and explain it:

Create account, error long

- Since we're talking about "user names", we should be consistent with the label.
- This is using PatternFly recommendations for errors in the forms, with the red helper text .
- I shortened the text to "User name must be shorter than 33 characters"
- Like the original mockups, the user name field is shorter, to show the limit visually. (It should roughly fit 33 wide characters, like 33 "w"s side by side, plus a little padding for the error icon.)

Create account, error reserved

- We don't need to say "please change" as that's implied
- I changed the text to be a little clearer
- I've added a (?) icon for a popover to explain it
- The popover shows the reserved words
- This icon and popover is a "nice to have" and should be a follow-up... do not block this PR for it :wink: 

Create account, error character type

- I changed the text to be shorter for the description and made it more consistent with the details (with the word first, then symbols after)

Again: Apologies for another review with different changes after you already addressed what I said yesterday. Thanks in advance!

I've also attached the changes as suggestions here, so I hope that makes it easier to apply. I did not see where the "User account" label is in this PR (so I could change it to "User name", to make it more consistent with the rest of the writing as well as with GNOME, KDE, and probably other desktops.)

rvykydal:
> 
>     * Since we're talking about "user names", we should be consistent with the label.
> 
>     * This is using PatternFly recommendations for errors in the forms, with the red helper text .
> 
>     * I shortened the text to "User name must be shorter than 33 characters"
> 
>     * Like the original mockups, the user name field is shorter, to show the limit visually. (It should roughly fit 33 wide characters, like 33 "w"s side by side, plus a little padding for the error icon.)
>

Modified. 33 "w"s seems maybe too wide to me (too close to the width of the other entries) but I am ok with it:

Screenshot from 2023-11-22 17-15-33

Screenshot from 2023-11-22 17-16-05




 

> 
>     * We don't need to say "please change" as that's implied
> 
>     * I changed the text to be a little clearer

Updated

> 
>     * I've added a (?) icon for a popover to explain it
> 
>     * The popover shows the reserved words
> 
>     * This icon and popover is a "nice to have" and should be a follow-up... do not block this PR for it 😉
> 

I'd leave it as eventual follow-up. 
(Personally I don't find the list very useful, IMO the change you hit the reserved name is quite low. Also there is a plan to move the validation to the backend so the backend would need to provide the list)

Screenshot from 2023-11-22 17-16-17




> 
>     * I changed the text to be shorter for the description and made it more consistent with the details (with the word first, then symbols after)
> 

Updated.
Just  a note - some rejected names falling under this message are not really consistent with the message (same as in gnome users where the original message comes from) but I think it is acceptable given they are pretty rare (maybe except for numbers-only). (https://github.com/rvykydal/anaconda-webui/blob/da5daf8533c5b030f8a78d36a0c1c8ec099aead7/src/components/users/Accounts.jsx#L81, https://github.com/rvykydal/anaconda-webui/blob/da5daf8533c5b030f8a78d36a0c1c8ec099aead7/test/check-users#L125)

Screenshot from 2023-11-22 17-15-48

garrett:
Thanks for the changes!

> Modified. 33 "w"s seems maybe too wide to me (too close to the width of the other entries) but I am ok with it:

Yeah, that does seem wide. I guess on a larger resolution screen, there might be more of a difference?

> I'd leave it as eventual follow-up.
> (Personally I don't find the list very useful, IMO the change you hit the reserved name is quite low. Also there is a plan to move the validation to the backend so the backend would need to provide the list)

Yeah, agreed on both points. It's pretty unlikely someone will hit multiples of the reserved words. Having it be a follow-up is totally fine. Forgetting to follow-up is probably OK too, as it's pretty unlikely.

> rejected names falling under this message are not really consistent with the message (same as in gnome users where the original message comes from) but I think it is acceptable given they are pretty rare (maybe except for numbers-only)

Should we have an error for numbers only then?

"User names must contain at least one letter from a-z."

Although I guess you could have a username of `$`, right? :thinking:

src/components/users/Accounts.jsx
```diff
@@ -55,6 +58,33 @@ export const accountsToDbusUsers = (accounts) => {
     }];
 };
 
+const reservedNames = [
```
KKoukiou, line 63: This code now is duplicated pyanaconda/core/users.py

I will create an issue to expose it via an API and remove the duplication.

KKoukiou (approved):

src/components/users/Accounts.jsx
KKoukiou: https://issues.redhat.com/browse/INSTALLER-3839

rvykydal:
> Thanks for the changes!
> 
> > Modified. 33 "w"s seems maybe too wide to me (too close to the width of the other entries) but I am ok with it:
> 
> Yeah, that does seem wide. I guess on a larger resolution screen, there might be more of a difference?
>

I don't think so, actually the width of other entries is already limited by limiting the form width: https://github.com/rvykydal/anaconda-webui/blob/da5daf8533c5b030f8a78d36a0c1c8ec099aead7/src/components/users/Accounts.scss#L3.
 

> > rejected names falling under this message are not really consistent with the message (same as in gnome users where the original message comes from) but I think it is acceptable given they are pretty rare (maybe except for numbers-only)
> 
> Should we have an error for numbers only then?
> 
> "User names must contain at least one letter from a-z."
> 
> Although I guess you could have a username of `$`, right? 🤔

If anything, I'd say something like "numbers-only user name is not allowed"  because for example those are valid: `3.4`, `4$`. (But I am fine with the current PR state)

KKoukiou (approved):

PR #48, opened 2023-11-20
webui: crypt user password by default before passing it to backend

rvykydal:
NOTE: using the same method as the current Gtk GUI

This is the PR https://github.com/rhinstaller/anaconda/pull/5321 migrated to the new repository.

### DIFF
--- a/src/components/users/Accounts.jsx
+++ b/src/components/users/Accounts.jsx
@@ -17,6 +17,8 @@
 
 import cockpit from "cockpit";
 import React, { useState, useEffect } from "react";
+import * as python from "python.js";
+import encryptUserPw from "../../scripts/encrypt-user-pw.py";
 
 import {
     Form,
@@ -45,12 +47,17 @@ export function getAccountsState (
     };
 }
 
+export const cryptUserPassword = async (password) => {
+    const crypted = await python.spawn(encryptUserPw, password, { err: "message", environ: ["LC_ALL=C.UTF-8"] });
+    return crypted;
+};
+
 export const accountsToDbusUsers = (accounts) => {
     return [{
         name: cockpit.variant("s", accounts.userAccount || ""),
         gecos: cockpit.variant("s", accounts.fullName || ""),
         password: cockpit.variant("s", accounts.password || ""),
-        "is-crypted": cockpit.variant("b", false),
+        "is-crypted": cockpit.variant("b", true),
         groups: cockpit.variant("as", ["wheel"]),
     }];
 };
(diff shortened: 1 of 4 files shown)


src/components/users/Accounts.jsx
```diff
@@ -45,12 +45,18 @@ export function getAccountsState (
     };
 }
 
+export const cryptUserPassword = async (password) => {
+    const pythonScript = `from random import SystemRandom as sr; import crypt; print(crypt.crypt("${password}", "$y$j9T$" + "".join(sr().choice("./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz") for _sc in range(24))))`;
```
KKoukiou, line 49: Please move this script in to scripts/myscript.py and importit.

You can then run it from JS with:
   import myscript from '../scripts/myscript.py';                                       

 python.spawn(myscript, args, { err: "message", environ: ['LC_ALL=C.UTF-8'] });

KKoukiou (changes requested):

KKoukiou (dismissed):

src/components/users/Accounts.jsx
```diff
@@ -45,12 +47,17 @@ export function getAccountsState (
     };
 }
 
+export const cryptUserPassword = async (password) => {
+    const crypted = await python.spawn(encryptUserPw, password, { err: "message", environ: ["LC_ALL=C.UTF-8"] });
```
KKoukiou, line 51: Actually, don't you want to handle the error in the front end? uncaught exceptions will result in critical error and finish the installation. I guess we are fine with that, but also leaving the password non crypted is not the end of the world. An notification should be enough.

rvykydal: Yes, it is on purpose. I think in such area (security) we should try to adhere to our policy/decision as much as possible. Also a notification would look a bit suspicious and puzzling to me as a user (like: hm, why are you crypting the password in the first place then, and what do I risk if I the password is not ecnrypted...). Also in the old UI we would fail hard with runtime error as well.

KKoukiou:
@rvykydal weirdly this consistently breaks the Progress test:
https://cockpit-logs.us-east-1.linodeobjects.com/pull-48-20231123-070254-c2bc29f0-fedora-rawhide-boot/log.html#8-2

rvykydal:
> @rvykydal weirdly this consistently breaks the Progress test: https://cockpit-logs.us-east-1.linodeobjects.com/pull-48-20231123-070254-c2bc29f0-fedora-rawhide-boot/log.html#8-2

oh yes, need to fix it:
```
Nov 23 07:31:32 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com org.fedoraproject.Anaconda.Modules.Users[4162]:   File "/usr/lib64/python3.12/site-packages/pyanaconda/core/users.py", line 503, in set_user_password
Nov 23 07:31:32 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com org.fedoraproject.Anaconda.Modules.Users[4162]:     raise OSError("Unable to set password for new user: status=%s" % proc.returncode)
Nov 23 07:31:32 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com org.fedoraproject.Anaconda.Modules.Users[4162]: OSError: Unable to set password for new user: status=1
```

KKoukiou (approved):

PR #50, opened 2023-11-20
Webui tests more robust going back

rvykydal:
This is port of PR https://github.com/rhinstaller/anaconda/pull/5310 to the new repository.
There is ongoing discussion in the original PR.

PR #51, opened 2023-11-20
Webui tests more robust going back and variant agnostic

rvykydal:
Port of https://github.com/rhinstaller/anaconda/pull/5331 to the new repository.

The first two commits are from this PR: https://github.com/rhinstaller/anaconda-webui/pull/50

With this PR, adding
```
pretend_live_iso(self, i)
```
(which effectively hides welcome and accounts step in the installer) before `open()` of all the tests where it makes sense the tests are still passing (except for pixel tests). So they should break less easily when adding some other step in the future (as it happened when I was adding users spoke).

As a follow-up the installer.py helper could be a bit cleaned and re-factored and maybe some tests run in multiple variants.

### DIFF
--- a/test/helpers/installer.py
+++ b/test/helpers/installer.py
@@ -40,6 +40,10 @@ class InstallerSteps(UserDict):
     _steps_jump[REVIEW] = PROGRESS
     _steps_jump[PROGRESS] = []
 
+    _parent_steps = {}
+    _parent_steps[DISK_ENCRYPTION] = DISK_CONFIGURATION
+    _parent_steps[CUSTOM_MOUNT_POINT] = DISK_CONFIGURATION
+
     _steps_callbacks = {}
     _steps_callbacks[ACCOUNTS] = create_user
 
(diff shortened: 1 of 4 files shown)


test/helpers/installer.py
```diff
@@ -40,6 +40,10 @@ class InstallerSteps(UserDict):
     _steps_jump[REVIEW] = PROGRESS
     _steps_jump[PROGRESS] = []
 
+    _parent_steps = {}
+    _parent_steps[DISK_ENCRYPTION] = DISK_CONFIGURATION
```
KKoukiou, line 44: We can get parent steps automatically. In fact we already do it.

previous_page = [k for k, v in self.steps._steps_jump.items() if current_page in v][0]

KKoukiou (approved):

test/helpers/installer.py
rvykydal: This gets previous page (INSTALLATION_METHOD), not parent page (DISK_CONFIGURATION). Parent page is the page/step with substeps which needs to be clicked on the sidebar to open the substeps and make them clickable.

PR #54, opened 2023-11-20
Update patternfly core to 5.2.0-prerelease.21

adamkankovsky:


package.json
```diff
@@ -44,7 +44,7 @@
   },
   "dependencies": {
     "@patternfly/patternfly": "5.1.0",
```
KKoukiou, line 46: The @patternfly group needs to be bumped in non-minor version updates all together. 

In other words you need to bump all 4 depedencies.

KKoukiou (changes requested):

KKoukiou (commented):
And please remove webui prefix.

package.json
adamkankovsky: I cant change a version of a react-log-viewer because most recent version is 5.1.0-prerelease.1

KKoukiou: you can try if the 'mix' of versions works or wait. Up to you.

KKoukiou (approved):

PR #58, opened 2023-11-21
tests: temporarily ignore in pixel tests the logo

KKoukiou:
This is not being always loaded.
https://issues.redhat.com/browse/INSTALLER-3792

Leave a single test in test/check-progress to check that icon till we figure out the issue.


rvykydal (approved):
Thank you.

PR #62, opened 2023-11-23
packit: Build COPR for main commits

KKoukiou:
Build every commit into main into
https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Anaconda-webui/

We will use that for cross-project testing (e.g. building ISOs from the main anaconda repo containing the anaconda-webui RPM with latest merged commit)

This now reuses the COPR project with the build after releases, but the latter are not used anywhere to this point, so no need for a new project.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -21,18 +21,12 @@ jobs:
     targets:
     - fedora-development
 
+  # for cross-project testing
   - job: copr_build
-    trigger: release
+    trigger: commit
     owner: "@rhinstaller"
     project: "Anaconda-webui"
     preserve_project: True
-    actions:
-      post-upstream-clone: make anaconda-webui.spec
-      # HACK: tarball for releases (copr_build, koji, etc.), copying spec's Source0; this
-      # really should be the default, see https://github.com/packit/packit-service/issues/1505
-      create-archive:
-        - sh -exc "curl -L -O https://github.com/rhinstaller/anaconda-webui/releases/download/${PACKIT_PROJECT_VERSION}/${PACKIT_PROJECT_NAME_VERSION}.tar.xz"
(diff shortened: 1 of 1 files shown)


packit.yaml
```diff
@@ -21,18 +21,13 @@ jobs:
     targets:
     - fedora-development
 
+  # for cross-project testing
   - job: copr_build
-    trigger: release
+    trigger: commit
+    branch: "^main$"
```
jkonecny12, line 27: I think we can omit this which will use default branch.

Honestly, I'm not sure if regular expressions works here...

jkonecny12 (approved):
Looks good to me otherwise.

VladimirSlavik (commented):
There's no branch and target specification, will that work?

KKoukiou:
> There's no branch and target specification, will that work?

I had it set before to main branch - but jirka preffered that I drop it. Indeed by default it picks up the 'Default branch'
https://packit.dev/docs/configuration/upstream/copr_build

VladimirSlavik:
And the default target is rawhide, I guess? What will happen if you branch? Or is that too far away to matter?

KKoukiou:
> And the default target is rawhide, I guess? What will happen if you branch? Or is that too far away to matter?

packit fetches targets from copr settings and updates them all. The copr settings for anaconda-webui is defined here: https://copr.fedorainfracloud.org/coprs/g/rhinstaller/Anaconda-webui/

There was fedora-39, fedora-eln and fedora-rawhide by default. I disabled the first two for now as we don't use these.

VladimirSlavik (approved):
OK, I guess it will work.

Personally I would prefer explicitly listing the branch and target, so that weird things do not happen once you have more of these, but deferring that to later makes sense too.

PR #63, opened 2023-11-23
tests: allow overriding the OS

KKoukiou:
We want to start testing webui against fedora ELN https://github.com/cockpit-project/bots/pull/5585

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,9 @@ COCKPIT_REPO_STAMP=pkg/lib/cockpit-po-plugin.js
 # stamp file to check if/when npm install ran
 NODE_MODULES_TEST=package-lock.json
 APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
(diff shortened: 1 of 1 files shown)


rvykydal (approved):

Makefile
```diff
@@ -27,7 +27,9 @@ COCKPIT_REPO_STAMP=pkg/lib/cockpit-po-plugin.js
 # stamp file to check if/when npm install ran
 NODE_MODULES_TEST=package-lock.json
 APPSTREAMFILE=org.cockpit-project.$(PACKAGE_NAME).metainfo.xml
+ifeq ($(TEST_OS),)
 TEST_OS=fedora-rawhide-boot
+endif
```
jkonecny12, line 32: Shouldn't we use rather 
```
TEST_OS ?= fedora-rawhide-boot
```

KKoukiou: It's the same result, feel free to adjust if you prefer.

PR #65, opened 2023-11-23
tests: generate screenshots with RTL language direction

KKoukiou:
This is keeping english and not some proper RTL language like Arabic but should be enough to identify overflows or other UI issues specific to this language group.

### DIFF
--- a/test/browser-layouts.json
+++ b/test/browser-layouts.json
@@ -2,6 +2,13 @@
     {
        "name": "desktop",
        "content_size": [1920, 1200],
+       "theme": "light",
        "shell_size": [1920, 1200]
+    },
+    {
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
LGTM!

jkonecny12:
I wonder, how the test cases know when to use RTL or standard images? I don't see any specification of that except the name in the json data.

KKoukiou:
> I wonder, how the test cases know when to use RTL or standard images? I don't see any specification of that except the name in the json data.

All tests cases will generate pixel tests automatically will generate also the RTL pixel test variant, by tweaking the browser configuration for the language direction on the fly.

PR #69, opened 2023-11-28
packit: add koji_build and bodhi_update targets

KKoukiou:
Trigger the build in Fedora Koji build system as a reaction to a new dist-git commit.
Create a new update in Fedora Bodhi for successful Koji build.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -38,3 +38,13 @@ jobs:
     trigger: release
     dist_git_branches:
       - fedora-development
+
+  - job: koji_build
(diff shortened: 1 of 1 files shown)


M4rtinK (approved):
Looks good to me, thanks! 

We do review the Packit distgit PRs, so automated builds afterwards make sense.

As for the Bodhi action - agreed on commenting it out for now. Updates should be created transparently on Rawhide. On branched Fedora, we can decide how to handle it once the time comes. :)

PR #72, opened 2023-11-30
storage: Add new integration layer between anaconda and cockpit storage

KKoukiou:
Blocked on:
- [x] https://github.com/rhinstaller/anaconda/pull/5422
- [x] https://github.com/cockpit-project/cockpit/pull/19883
- [x] https://github.com/cockpit-project/cockpit/pull/19876
- [x] https://issues.redhat.com/browse/INSTALLER-3880
- [x] LUKS spport in the new scenario - TODO katerina this PR
- [x] Rescanning deactivates LVM volumes - Marius introduced a workaround where c-storage handles inactive volumes
- [x] Btrfs full support https://github.com/cockpit-project/cockpit/pull/19880 https://github.com/cockpit-project/cockpit/pull/19872
- [ ] Set "default_fsys_type" in "cockpit_anaconda" when starting Cockpit Storage.
- [x] Merge https://github.com/cockpit-project/cockpit/pull/19924

### DIFF
--- a/src/components/storage/CockpitStorageIntegration.jsx
+++ b/src/components/storage/CockpitStorageIntegration.jsx
@@ -49,6 +49,9 @@ import {
 } from "../../apis/storage.js";
 
 import {
+    unlockDevice,
+} from "../../apis/storage_devicetree.js";
+import {
     setBootloaderDrive,
 } from "../../apis/storage_bootloader.js";
 import {
@@ -221,6 +224,7 @@ const CheckStorageDialog = ({
     const requiredSize = useRequiredSize();
 
     const newMountPoints = useMemo(() => JSON.parse(window.sessionStorage.getItem("cockpit_mount_points") || "{}"), []);
+    const cockpitPassphrases = useMemo(() => JSON.parse(window.sessionStorage.getItem("cockpit_passphrases") || "{}"), []);
 
     const useConfiguredStorage = useMemo(() => {
         const availability = checkConfiguredStorage({
@@ -246,6 +250,52 @@ const CheckStorageDialog = ({
             setCheckStep();
         }
     }, [useConfiguredStorage, checkStep]);
+
+    useEffect(() => {
+        if (checkStep !== "luks") {
+            return;
+        }
+
+        const cockpitDevices = (
+            Object.keys(newMountPoints)
+                    .map(devicePath => ({
+                        devicePath,
+                        deviceName: getDeviceNameByPath(deviceData, devicePath),
+                    }))
+        );
+
+        const devicesToUnlock = (
+            cockpitDevices
+                    .filter(({ devicePath, deviceName }) => {
+                        return (
+                            newMountPoints[devicePath].type === "crypto" &&
+                            deviceData[deviceName].formatData.attrs.v.has_key !== "True"
+                        );
+                    })
+                    .map(({ devicePath, deviceName }) => ({
+                        deviceName,
+                        passphrase: cockpitPassphrases[devicePath],
+                    }))
+        );
+
+        if (devicesToUnlock.some(dev => !dev.passphrase)) {
+            onCritFail()({ message: _("Cockpit storage did not provide the passphrase to unlock encrypted device.") });
+        }
+
+        if (devicesToUnlock.length === 0) {
+            setCheckStep("prepare-partitioning");
+            return;
+        }
+
+        Promise.all(devicesToUnlock.map(unlockDevice))
+                .catch(exc => {
+                    setCheckStep();
+                    setError(exc);
+                })
+                .then(() => {
+                    dispatch(getDevicesAction());
+                });
+    }, [dispatch, checkStep, cockpitPassphrases, newMountPoints, deviceData, onCritFail, setError]);
 
     useEffect(() => {
         // If the required devices needed for manual partitioning are set up,
@@ -291,7 +341,7 @@ const CheckStorageDialog = ({
                         onSuccess: () => dispatch(getDevicesAction())
                                 .then(() => {
                                     if (useConfiguredStorage) {
-                                        setCheckStep("prepare-partitioning");
+                                        setCheckStep("luks");
                                     } else {
                                         setCheckStep();
                                     }
--- a/src/components/storage/CockpitStorageIntegration.jsx
+++ b/src/components/storage/CockpitStorageIntegration.jsx
@@ -16,17 +16,21 @@
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
 import cockpit from "cockpit";
-import React, { useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 
 import {
+    ActionList,
     Alert,
     Button,
     Card,
     CardBody,
     Flex,
     FlexItem,
+    HelperText,
+    HelperTextItem,
     List,
     ListItem,
+    Modal,
     PageSection,
     PageSectionVariants,
     Text,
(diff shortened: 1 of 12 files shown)


src/components/storage/InstallationDestination.jsx
```diff
@@ -433,7 +433,7 @@ export const InstallationDestination = ({
                                 : _("No usable disks detected")
                         )}
                     {rescanDisksButton}
-                    {!isBootIso && <ModifyStorage idPrefix={idPrefix} onCritFail={onCritFail} onRescan={onClickRescan} />}
```
mvollmer, line 436: I had to remove the `isBootIso` check only so that I could manually test the "Modify storage" button, since it the flag was true in my test setup. I don't understand at all what it is about, so removing this check as part of the Cockpit integration is probably wrong.

src/components/storage/MountPointMapping.jsx
```diff
@@ -58,6 +58,7 @@ import {
 import "./MountPointMapping.scss";
 
 const _ = cockpit.gettext;
+const cmpm = JSON.parse(window.localStorage.getItem("cockpit_mount_points"));
```
mvollmer, line 61: Cockpit is no longer exporting this information.

src/components/storage/ModifyStorage.jsx
```diff
@@ -29,19 +29,24 @@ import { WrenchIcon, ExternalLinkAltIcon } from "@patternfly/react-icons";
...
-                return onStarted();
-            })
-            .catch((error) => { window.clearTimeout(timeoutId); errorHandler(error) });
+let cockpit_window = null;
+
+const startCockpitStorage = (onStart, onStarted, errorHandler) => {
+    window.localStorage.setItem("cockpit_anaconda",
+                                JSON.stringify({
+                                    mount_point_prefix: "/mnt/sysimage",
+                                    ignore_devices: ["/dev/sr0", "/dev/loop0"]
```
mvollmer, line 38: This has changed to `available_devices` and Anaconda could probably easily fill this now with real information from the backend.

mvollmer:
@KKoukiou, could you rebase this? Thanks!

mvollmer:
I made a version of this that works with https://github.com/cockpit-project/cockpit/pull/19352: https://github.com/mvollmer/anaconda-webui/tree/cockpit-storage

https://github.com/cockpit-project/cockpit/pull/19352 will not change incompatibly any more.

mvollmer:
I have pushed my version here.

rvykydal:
I am seeing an issue: if I go to "Mount point assignment", then back and then to "Mount point assignment" again, required and recommended rows disappear. I hit it for example after creating one 10G ext4 partition (created in shell or via cockpit-storage). Rescan before going to the "Mount point assignment" makes the required/recommendend mount points to appear again.

src/components/storage/InstallationScenario.jsx
```diff
@@ -92,7 +99,56 @@ const checkMountPointMapping = ({ hasFilesystems, duplicateDeviceNames }) => {
...
+                            }
+                        };
+
+                        if (m["mount-point"].v) {
+                            Object.keys(newMountPoints).forEach(key => getNestedDirs(newMountPoints[key]));
+
+                            return allDirs.includes(m["mount-point"].v);
+                        }
+
+                        // FIXME: cockpit does not return the biosboot partitions in the cockpit_mount_points
```
mvollmer, line 138: Why would it? A biosboot partition has no mount point. Doesn't blivet find them based on their partition type?

src/components/storage/ModifyStorage.jsx
```diff
@@ -41,12 +41,12 @@ export const ModifyStorage = ({ idPrefix, onCritFail, onRescan, setShowStorage,
...
-                                              JSON.stringify({
-                                                  mount_point_prefix: targetSystemRoot,
-                                                  available_devices: selectedDevices,
-                                                  efi: isEfi,
-                                              })
+                  window.sessionStorage.setItem("cockpit_anaconda",
+                                                JSON.stringify({
+                                                    mount_point_prefix: targetSystemRoot,
+                                                    available_devices: selectedDevices,
+                                                    efi: isEfi,
```
mvollmer, line 48: This should also set `"default_fsys_type"`.

test/check-storage
```diff
@@ -1047,6 +1048,217 @@ class TestStorageMountPoints(anacondalib.VirtInstallMachineCase, StorageCase):
...
+        self.dialog({"size": 1070, "type": "ext4", "mount_point": "/boot"})
+
+        self.click_dropdown(self.card_row("Storage", 4), "Create partition")
+        self.dialog({
+            "type": "ext4", "mount_point": "/",
+            "crypto": self.default_crypto_type,
+            "passphrase": "redhat",
+            "passphrase2": "redhat",
+        })
+        # FIXME: Cockpit should leave open LUKS devices afetr creation
```
mvollmer, line 1090: See https://github.com/cockpit-project/cockpit/pull/20016

src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -247,6 +251,52 @@ const CheckStorageDialog = ({
...
+                        );
+                    })
+                    .map(({ devicePath, deviceName }) => ({
+                        deviceName,
+                        passphrase: cockpitPassphrases[devicePath],
+                    }))
+        );
+
+        if (devicesToUnlock.some(dev => !dev.passphrase)) {
+            onCritFail()({ message: _("Cockpit storage did not provide the passphrase to unlock encrypted device.") });
```
mvollmer, line 282: Wouldn't Anaconda need to prompt the user for the passphrase in this case? It is always possible that some LUKS device is closed for some reason.

src/components/storage/ModifyStorage.jsx
KKoukiou: I really dont know where to parse it from. I dont think there is an API for that currently. I prefer to figure it out in followup.

src/components/storage/CockpitStorageIntegration.jsx
KKoukiou: I agree, that this can be improved. However, since that more of an edge condition, I would suggest to implement the unlock code plus test for it as followup.

src/components/storage/InstallationScenario.jsx
KKoukiou: Yes I will remove the comment.

garrett:
Hi, I managed to finally build this PR again. (It takes between 10 to 20 minutes here on my lowly laptop, when it works... and there were a few issues with stale builds and images.)

Anyway, configure storage is still not implemented as a modal (not a dialog styled modal, but a larger modal). For reference, this is how it should look:

image

Part of the idea of the original design was to have it as an overlay like the above mockup (made from sandwiching screenshots together) so that it doesn't look like part of Anaconda's normal flow. This is intended to look like a "separate app" so people would understand that it's doing something else. It helps with the mental modal of doing a separate task to prepare the system.

(We actually implemented this for blivet-gui for F39, but as a modal that showed up behind blivet-gui, which gave it the same effect. This time around, it should just show the backdrop, not the full modal. But it's important to have this similar layer to look like a separate app.)

Having it take over 100% of the screen makes it look like it's still part of Anaconda, which is causing some confusion.

We could also really change the top area to be more prominent. The heading is kind of redundant anyway, so if we outright replace it with an alert at the top, it could look more like this:

image


But this can be implemented as a follow-up. I think it's important to merge what we have here already and address other issues as follow-ups.

src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -247,6 +251,52 @@ const CheckStorageDialog = ({
...
+        }
+
+        if (devicesToUnlock.length === 0) {
+            setCheckStep("prepare-partitioning");
+            return;
+        }
+
+        Promise.all(devicesToUnlock.map(unlockDevice))
```
mvollmer, line 290: `cockpit_mount_points` does not contain non-filesystem things like LVM2 physical volumes, so if you make a encrypted LVM2 physical volume, `cockpit_passprhases` will contain the passphrase for it, but Anaconda will not unlock it since it isn't also mentioned in `cockpit_mount_points`. What about just unlocking everything that is mentioned in `cockpit_passphrases`, without filtering out devices that are not mentioned in `cockpit_mount_points`?

src/components/storage/CockpitStorageIntegration.jsx
mvollmer: I just hacked it to be like this and that works fine:

```
    useEffect(() => {
        if (checkStep !== "luks") {
            return;
        }

        const devicesToUnlock = (
            Object.keys(cockpitPassphrases).map(dev => ({ deviceName: getDeviceNameByPath(deviceData, dev),
                                                          passphrase: cockpitPassphrases[dev]
                                                        })));

        console.log("devicesToUnlock", devicesToUnlock);

        Promise.all(devicesToUnlock.map(unlockDevice))
                .catch(exc => {
                    setCheckStep();
                    setError(exc);
                })
                .then(() => {
                    setCheckStep("prepare-partitioning");
                });
    }, [dispatch, checkStep, cockpitPassphrases, newMountPoints, deviceData, onCritFail, setError]);
```

src/components/storage/CockpitStorageIntegration.jsx
KKoukiou: This probably wont work if you exist storage mode and enter again, as it will try to unlock unlocked devices. But I see the approach, will adjust.

mvollmer:
Nice!

I keep thinking that as a user, I would like to see the summary of what is being done immediately after "returning to installation", to double check whether Anaconda has correctly understood what I want to do. Now I have to fill out the account information before I can see the summary. Thus, the account stuff gets "in the way" of finishing the storage configuration.

KKoukiou:
> Nice!
> 
> I keep thinking that as a user, I would like to see the summary of what is being done immediately after "returning to installation", to double check whether Anaconda has correctly understood what I want to do. Now I have to fill out the account information before I can see the summary. Thus, the account stuff gets "in the way" of finishing the storage configuration.

I agree. @garrett can you give this some thought?

garrett:
Sure, that makes sense. I thought Cockpit doesn't track the changes, however, which prevented us from showing this?

KKoukiou:
> Sure, that makes sense. I thought Cockpit doesn't track the changes, however, which prevented us from showing this?
> 
> I wanted that too, but was told that Cockpit doesn't track the changes and Anaconda doesn't know about the changes either. So I don't know how we could even get this information or what it would be like.
> 
> Plus, at this stage, there is nothing to be done, as it was already done. Cockpit, via udisks, doesn't queue up changes, but acts on them immediately.

All the information we present in the review page, (which partitions have which mount points etc) are already present  when that dialog appears that allows us to exit storage mode.

We can add these information in that dialog if you like.

garrett:
When I go to create a biosboot partition, it wants to give it all the available space. Why is space even configurable for this? Isn't it supposed to be 1 megabyte. Why does it have overwrite as an option too?

image

garrett:
image

It says "Apply storage configuration and install". But didn't we already do storage configuration in Cockpit? Shouldn't it just be "Install" at this point in time?

src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -121,15 +132,307 @@ export const CockpitStorageIntegration = ({
...
+                    setError(exc);
+                })
+                .then(() => {
+                    dispatch(getDevicesAction());
+                });
+    }, [dispatch, checkStep, cockpitPassphrases, newMountPoints, deviceData, onCritFail, setError]);
+
+    useEffect(() => {
+        // If the required devices needed for manual partitioning are set up,
+        // prepare the partitioning and set the scenario to 'use-configured-storage'.
```
rvykydal, line 294: I can't find the "set the scenario to 'use-configured-storage' part.

rvykydal (approved):
Looks good to me, thank you.
Brave work!

src/components/storage/CockpitStorageIntegration.jsx
KKoukiou: Indeed, code was refactored. removed.

PR #76, opened 2023-12-04
Fix up some docs and infra files to reflect the repo structure

KKoukiou:
Fallouts after the repo split.


adamkankovsky (approved):
Thanks for correction

PR #90, opened 2023-12-07
Adjust error reporting dialog for RHEL

adamkankovsky:
This PR is just a draft for now because there are a lot of things that need to be fixed before it can be finished. 

The first one is a blocker, because of which it is not possible to insert your own description. Jira issue describing this problem: https://issues.redhat.com/browse/RIJ-612

The second thing is that it is not yet known how the os-release will look like in RHEL10. Currently, the information needed for error reporting related to the bugzilla is stored in the os-release. I've already pointed this out and it seems to be actively addressed by rhel-devel. I will continue to monitor the situation.


KKoukiou:
Let's reopen when it's actively worked on. The branch will be there.

PR #97, opened 2023-12-11
tests: explicitely specify the COPR repo for downloading the RPMs if …

KKoukiou:
…on -pr- scenario

The dnf will prefer the package from rawhide if the version is higher than the one available in the PRs COPR.


rvykydal (approved):
Maybe the `dnf download` (together with `dnf copr enable`) could go to some helper function now, like `download_copr_build` but I don't insist, maybe next time.

KKoukiou:
This is blocked on https://github.com/rhinstaller/anaconda/pull/5376

unfortunately we were not really installing from the copr till this PR.

PR #98, opened 2023-12-12
Create copr builds and such only on the rawhide now

KKoukiou:
Packit thinks that fedora-development is also f39 and f-rawhide, it starts to build the anaconda webui RPM in f39 and as expected this does not work.

Let's make the target more explicit.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -19,7 +19,7 @@ jobs:
   - job: copr_build
     trigger: pull_request
     targets:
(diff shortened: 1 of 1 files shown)


KKoukiou:
Looks like it got resolved by itself. Packit built just rawhide copr.

VladimirSlavik:
I think the change was correct, unless you want to build for the released fedora too, from the one branch?

PR #100, opened 2023-12-14
tests: execute anaconda specific cleanup handlers after the cockpit's ones

KKoukiou:
Cockpit's cleanup handlers are responsible among others for cleaning up per-test added disks, such as SCSI disks. We need to invoke the parent class setUp after adding the anaconda specific cleanup handers.
This sequence guarantees that re-scanning occurs at the very end of the cleanup process, after all disks are removed.

This was causing problems in CI and regressed with 899d47eb8d22ed2bfa52c215dbec31fa7efe1abf.


velezd (approved):

velezd:
Looks like `super().setUp()` needs to be moved above `self.allow_journal_messages`

PR #105, opened 2023-12-18
DEBUG: debug timeout on s.rescan_disks

rvykydal:


martinpitt:
Note: the "vkCreateInstance: Found no drivers!" errors from [this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-95-20240104-142225-af47dfc6-fedora-eln-boot/log.html#33) happen in [other projects, too](https://cockpit-logs.us-east-1.linodeobjects.com/pull-1370-20240104-150421-edbe2c28-centos-8-stream/log.html#79-2). They are quite bogus -- this is *headless* chrome, it shouldn't have any business with even *trying* to do anything with graphics.

But [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-105-20240104-141808-c82be553-fedora-eln-boot/log.html#33) is more serious -- the browser actually crashed, which looks like the root cause for the failing wait_not_present() and screenshot and the "ECONNREFUSED 127.0.0.1:10012" (which is presumably the CDP port).

I have no off-hand idea or workaround. @KKoukiou says this doesn't reproduce locally, so I'll try to repro on our production bots tomorrow and see if I'll find anything.

martinpitt:
Switching to Firefox in PR #119 is the better approach here for sure -- let's rather debug problems with that.

However, I gave this some time-boxed investigation out of curiosity. In a production tasks container, I did
```
git clone https://github.com/rhinstaller/anaconda-webui/
cd anaconda-webui
export TEST_OS=fedora-eln-boot
make prepare-test-deps test/reference updates.img 
TEST_AUDIT_NO_SELINUX=1 test/common/run-tests TestStorageMountPointsEFI.testBasic
```

This fails immediately with "FileNotFoundError: Can't run tests without a prepared image; use test/image-prepare" -- huh? that's not how anaconda works. So I modified `Makefile` to only run that single test, and ran `TEST_OS=fedora-eln-boot make integration-test` instead.  But that also passes -- it seems it may fail only when it uses more resources, i.e. more parallel browsers/VMs?

So I started the full test run, with `watch df -h` and `top` running. There are a lot of "legit" test failures, similar to the ones I saw [with firefox](https://github.com/rhinstaller/anaconda-webui/pull/119#issuecomment-1878218870), but no browser crashes. However, it did use quite a lot of memory in between, 30% out of the 128 GiB. /dev/shm never used significantly much (it topped at 7% out of the 1 GiB or so).

So my theory is, this is likely to happen when our CI machines are busy, and the other tasks containers run a lot of tests as well. We probably have to reduce parallelism a bit to avoid running out of memory. I don't see any OOM related things in ci-srv-06's journal for that test run, but given how much RAM this uses, it's at least a plausible theory -- and it's corroborated by @KKoukiou 's inability to reproduce this locally (where you don't have a zillion parallel tests running).

martinpitt:
@KKoukiou so if you want to test this, maybe add something like this to test/run, to see if it helps?
```sh
[ ${TEST_JOBS:-1} -le 4 ] || export TEST_JOBS=4
```

martinpitt:
... and if it does, we can see how we teach that to the infra.

PR #119, opened 2024-01-04
tests: run CI against firefox only

KKoukiou:
Chromium is the default browser for cockpit based tests but we care only about firefox at this point.

### DIFF
--- a/test/browser-layouts.json
+++ b/test/browser-layouts.json
@@ -1,6 +1,7 @@
 [
     {
        "name": "desktop",
+       "theme": "light",
        "content_size": [1920, 1200],
        "theme": "light",
        "shell_size": [1920, 1200]
--- a/test/run
+++ b/test/run
@@ -5,4 +5,10 @@
 set -eux
 
 make codecheck
+
+# The default and only available browser inside the installation environment is firefox so even
+# though we are testing remote installation we are closer to the real use case by testing against
+# firefox
+export TEST_BROWSER=firefox
+
 make integration-test


martinpitt:
This [needs a little tweak](https://cockpit-logs.us-east-1.linodeobjects.com/pull-119-20240104-154124-43dcfc01-fedora-eln-boot/log.html). I took the liberty to push it.

martinpitt:
Locally, a `make integration-tests` run doesn't get very far. `TestBasic.testAboutModal` passes, but the next one `TestBasic.testErrorHandling` times out. It does the same on CI, the retry works.

Otherwise it runs on CI in general -- both the [ELN](https://cockpit-logs.us-east-1.linodeobjects.com/pull-119-20240105-061849-f182aa75-fedora-eln-boot/log.html) and the [rawhide](https://cockpit-logs.us-east-1.linodeobjects.com/pull-119-20240105-061849-f182aa75-fedora-rawhide-boot/log.html) runs have "legit" failures which are not browser specific at first sight -- e.g. [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-119-20240105-061849-f182aa75-fedora-rawhide-boot/log.html#32) the "Next" button is disabled, and [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-119-20240105-061849-f182aa75-fedora-eln-boot/log.html#32) the password strength says "Medium" instead of the expected "strong". The two TestStorageMountPoints failures just look like timeouts, as it's still detecting disks. Bumping the timeout might work wonders there? Or rather, waiting for the initialization to be finished, with a 60s timeout -- I don't know what to wait for, but @KKoukiou probably does. I can reproduce the same with chromium, it seems to always work on the second attempt. It's also the common theme on the [weather report](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?days=7&repo=rhinstaller%2Fanaconda-webui).

However, tests are rather slow (many timeouts) and flaky, much more than with chromium. So this isn't ready just yet until we know the reason for the timeouts.

KKoukiou:
The testReportIssue and testErrorHandling fail on the same things that's 100% reproduceable locally. On firefox Ctrl+a + Backspace + Type new string freezes at some point, while on of these actions are done on the text area. 

* At start i thought there is some issue with the textarea hosting a big size input, this is not the problem, I trimmed the journal before starting the test so that it's < 100 lines and the issue persisted
* Then I thought textarea might be the problem, so I tried to temporarily switch to TextInput, but the same problem persisted.
* Note that this problem happens only on headless firefox locally, the interactive one works as expected.

PR #120, opened 2024-01-05
Root account

rvykydal:
TODO:
- [x] Add tests for password setting
- [ ] Create follow-up issue(s) (SSH login, pw validation, review screen update, check status review, strength indicator improvements...)
Screenshot from 2024-01-05 13-05-50
Screenshot from 2024-01-05 13-06-13

### DIFF
--- a/src/components/users/Accounts.jsx
+++ b/src/components/users/Accounts.jsx
@@ -20,16 +20,23 @@ import React, { useState, useEffect } from "react";
 import * as python from "python.js";
 import encryptUserPw from "../../scripts/encrypt-user-pw.py";
 import { debounce } from "throttle-debounce";
+import {
+    setUsers,
+    setIsRootAccountLocked,
+    setCryptedRootPassword,
+    clearRootPassword,
+} from "../../apis/users.js";
 
 import {
+    Checkbox,
     Form,
     FormGroup,
     FormHelperText,
+    FormSection,
     HelperText,
     HelperTextItem,
     InputGroup,
     TextInput,
-    Title,
 } from "@patternfly/react-core";
 
 import "./Accounts.scss";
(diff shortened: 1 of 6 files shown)


src/components/users/Accounts.jsx
```diff
@@ -43,12 +50,18 @@ export function getAccountsState (
     userName = "",
     password = "",
     confirmPassword = "",
+    rootPassword = "",
+    rootConfirmPassword = "",
+    enableRoot = false,
```
rvykydal, line 55: Maybe I should rather use rootEnabled or isRootEnabled here.

src/components/users/Accounts.jsx
```diff
@@ -235,6 +262,69 @@ const CreateAccount = ({
...
+          isHorizontal
+          id={idPrefix}
+        >
+            <Title
+              headingLevel="h2"
+              id={idPrefix + "-title"}
+            >
+                {_("System")}
+            </Title>
+            <Form>
```
rvykydal, line 321: This <Form> seems to be superfluous..

src/components/users/Accounts.jsx
KKoukiou: Self review is the best review :P

KKoukiou: I like isRootEnabled :)

src/components/users/Accounts.jsx
KKoukiou: Right, you want one form per page. Remove the <Form> wrapper from the CreateAccount component, and put it in the Accounts component so that it wraps also your newelly added one.

src/components/users/Accounts.jsx
```diff
@@ -235,6 +262,69 @@ const CreateAccount = ({
...
+    useEffect(() => {
+        setAccounts(ac => ({ ...ac, rootPassword: password, rootConfirmPassword: confirmPassword }));
+    }, [setAccounts, password, confirmPassword]);
+
+    return (
+        <Form
+          isHorizontal
+          id={idPrefix}
+        >
+            <Title
```
KKoukiou, line 315: I think the propper approach would be to user FormSection component. https://www.patternfly.org/components/forms/form#form-sections

KKoukiou (commented):
Two minor improvements and it's good to go.

src/components/users/Accounts.jsx
rvykydal: Updated

src/components/users/Accounts.jsx
rvykydal: Used, it produces slightly different layout in tests (spacing of the sections and title sizes) but for me it is acceptable.

rvykydal:
I updated the <Form> using.
I also added basic validation rule (length) which brings the current checking on the same level as user password. I'll create another issue for the validation review (compared to Gtk frontend), we seem to be missing ascii warning (in webui it is present only in storage encryption) but most importantly we are ignoring checks performed by the tool using libpwquality.

KKoukiou (approved):

garrett (approved):
Looks good for the context of this PR!

Please work on follow-ups to:
- fix the error and success strings: "Passwords do not match" / "Passwords match" (using "Passwords must match" for a success state string indicates that they do not match.)
- improve the password check
  - show a "-" icon by default; do not actively check until one or the other happens: 1. string length is the same between password and repeat password 2. you focus away from the input (aka: blur()). When one of these things happen, then check the string with a pass/fail event and again on every additional keypress when in the form.

Additionally:
- we should fix the strength indicator; it's in the wrong place and doesn't indicate what it is ("Medium" and "Strong" mean nothing when they're the only word) — since PatternFly doesn't handle this, I think you and I probably need to work on that together.

For context on this, Cockpit does the following:

Accounts Create

It has a password strength indicator with a strength meter under the entry

Accounts Create(1)

It does the "Passwords do not match" check as indicated above.

PR #132, opened 2024-01-16
Use flag file to get signaled when backend is ready

KKoukiou:


src/components/app.jsx
```diff
@@ -62,6 +63,44 @@ export const Application = () => {
...
+        const channel = cockpit.channel({
+            payload: "fswatch1",
+            path: "/run/anaconda/",
+            superuser: "try",
+        });
+
+        channel.addEventListener("message", (ev, data) => {
+            const item = JSON.parse(data);
+
+            if (item.event === "created" && item.path === "/run/anaconda/backend_ready") {
```
jelly, line 90: if I recall correctly you can use `cockpit.file("/run/anaconda/backend_ready").watch()` directly on a non-existing file.

KKoukiou: I was wondering that's why I asked you to review :P ok I will check!

KKoukiou: Works perfectly thanks

jkonecny12:
So I was closer with my solution than I tough. That is great!

Thanks for helping me. I'll test it out.

rvykydal (approved):

PR #134, opened 2024-01-21
Exclude Firefox on ELN

jkonecny12:
We can't use Firefox in ELN because it was removed. This Requires is dragging it back.

https://github.com/minimization/content-resolver-input/pull/1061


jkonecny12:
@KKoukiou could we please get this in and make a release soon?

KKoukiou (commented):
The change itself is ok for now. Don't we need to include though flatpak package instead in RHEL as hard requirement? 
Also as talked on the chat we need:
* Jira ticket on our side to switch the firefox spawning to flatpack `flatpak run org.mozilla.firefox`
* Jira ticket on image builder team to include the flatpak in the ISOs or whatever is needed that

jkonecny12:
I'm not aware of any way how we could have requirement on Flatpak from RPM package. We can just check the binary runtime as far as I know.

KKoukiou (approved):

PR #145, opened 2024-01-29
Use the "encryption-allowed" constraint in UI checks

KKoukiou:


KKoukiou:
@rvykydal I split this out from https://github.com/rhinstaller/anaconda-webui/pull/140 to unblock the commits that dont fail.

PR #152, opened 2024-02-01
Add early biosboot check to mountpoint assignment

rvykydal:


AdamWill:
"Some of required partitions are missing" is not good English. "Some required partitions are missing" would be better.

rvykydal:
> Thanks. It still feels wrong to me that we're essentially duplicating all these checks between the backend and the UI layer - in my experience, having different code in different places doing the same thing is a recipe for future bugs. But I guess it's still not feasible to refactor things so the existing backend checks can be used in the frontend here?

I definitely don't want to duplicate the checks, that is why I am adding only static information (independent of the current assignment requests) the backend is able to provide at this point (MountPointConstraints API). Using the backend checks dynamically on the UI update (in case of mount point assignment, probably easier) or storage update (in case of cockpit-storage, requiring support in blivet) is beyond the scope of this PR.

Also it would be helpful to have them written-down as concrete use cases to be addressed (similar as in the BZ).

Regarding duplication in this PR, I think this information https://github.com/rvykydal/anaconda-webui/blob/7d7188e17975272dd437f673507bfb8f5dd8f584/src/components/storage/InstallationScenario.jsx#L80 should not be part of the UI but passed from backend in the constraints.

rvykydal:
> Regarding duplication in this PR, I think this information https://github.com/rvykydal/anaconda-webui/blob/7d7188e17975272dd437f673507bfb8f5dd8f584/src/components/storage/InstallationScenario.jsx#L80 should not be part of the UI but passed from backend in the constraints.

Can be obtained from the backend:  https://github.com/rhinstaller/anaconda/blob/a721090e987994a947245aebc31dfec0e1927ac6/pyanaconda/modules/storage/devicetree/viewer_interface.py#L85, I'll update the PR.
(For the record there is also some further detailed info in Gtk frontend: https://github.com/rhinstaller/anaconda/blob/a721090e987994a947245aebc31dfec0e1927ac6/pyanaconda/core/storage.py#L106)

rvykydal:
> I definitely don't want to duplicate the checks, that is why I am adding only static information (independent of the current assignment requests) the backend is able to provide at this point (MountPointConstraints API). Using the backend checks dynamically on the UI update (in case of mount point assignment, probably easier) or storage update (in case of cockpit-storage, requiring support in blivet) is beyond the scope of this PR.
> 

It has to deal with the fact that most of the checks (be it more granular checks in the interactive/custom paritioning or the batch test run in ApplyParitioning) require the whole partitioning model (therefore its update in the backend on any particular update in UI that should be checked). But perhaps it should be feasible for mount point assignment, where it means recreating the partitioning (model) from the requests via the API (which is currently done on Next) and doing the partitioning validation check on the partitioning.

rvykydal:
> "Some of required partitions are missing" is not good English. "Some required partitions are missing" would be better.

Fixed

KKoukiou (approved):

KKoukiou (commented):
The failure in the testMultiple disks is not caused by this PR - it's flake and I am trying to address it.

src/components/storage/Common.jsx
```diff
@@ -72,22 +73,20 @@ export const useDuplicateDeviceNames = ({ deviceNames }) => {
...
-            selectedDisks.some(device => (
-                devices[device]?.children.v.some(child => (
-                    devices[child]?.formatData.mountable.v || devices[child]?.formatData.type.v === "luks")
-                )
-            ))
-        );
-
-        setHasFilesystems(_hasFilesystems);
+        const _usablePartitions = Object.values(devices).filter(device =>
+            (device.formatData?.type.v === "biosboot" ||
```
rvykydal, line 81: Oh, I have just noticed, having the string here does not seem right.
It should be more generic check or it should at least include also "prepboot" and "appleboot".
Something for a potential follow-up I think.

rvykydal:
Looking into the TestStorageMountPoints.testLVM issue (TestStorageMountPoints.testLVM). Not reproducible when running a single test, I suspect counting rows here:https://github.com/rhinstaller/anaconda-webui/blob/8a36e907a54c05fb8e2502aec6f4c6414402e03a/test/helpers/storage.py#L491 too early.

PR #164, opened 2024-02-07
packaging: fix double Requires of cockpit-storaged in the spec file

KKoukiou:
Also it's not cockpit-storage but cockpit-storaged.

Remove obsolete comment.


jkonecny12 (approved):
LGTM!

PR #167, opened 2024-02-08
Hostname handling

adamkankovsky:
image
image

### DIFF
--- a/src/components/review/ReviewConfiguration.jsx
+++ b/src/components/review/ReviewConfiguration.jsx
@@ -29,8 +29,9 @@ import {
 } from "@patternfly/react-core";
 
 import { AnacondaWizardFooter } from "../AnacondaWizardFooter.jsx";
-import { FooterContext, LanguageContext, OsReleaseContext, StorageContext, UsersContext } from "../Common.jsx";
+import { FooterContext, LanguageContext, OsReleaseContext, StorageContext, SystemTypeContext, UsersContext } from "../Common.jsx";
 import { useScenario } from "../storage/InstallationScenario.jsx";
+import { HostnameRow } from "./Hostname.jsx";
 import { StorageReview } from "./StorageReview.jsx";
 
 import "./ReviewConfiguration.scss";
@@ -61,6 +62,7 @@ const ReviewConfiguration = ({ idPrefix, setIsFormValid }) => {
     const accounts = useContext(UsersContext);
     const { partitioning, storageScenarioId } = useContext(StorageContext);
     const { label: scenarioLabel } = useScenario();
+    const isBootIso = useContext(SystemTypeContext) === "BOOT_ISO";
 
     useEffect(() => {
         setIsFormValid(true);
@@ -118,6 +120,10 @@ const ReviewConfiguration = ({ idPrefix, setIsFormValid }) => {
                             </DescriptionListDescription>
                         </DescriptionListGroup>
                     </ReviewDescriptionList>
+                    {isBootIso &&
+                    <ReviewDescriptionList>
+                        <HostnameRow />
+                    </ReviewDescriptionList>}
                 </ReviewDescriptionList>
             </FlexItem>
             <FlexItem>
--- a/src/components/review/ReviewConfiguration.jsx
+++ b/src/components/review/ReviewConfiguration.jsx
@@ -21,6 +21,7 @@ import {
     Button,
     DescriptionList, DescriptionListDescription,
     DescriptionListGroup, DescriptionListTerm,
+    Flex, FlexItem,
     HelperText, HelperTextItem,
     Modal, ModalVariant,
     Stack,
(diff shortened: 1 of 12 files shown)


src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
     );
 };
 
-export const ReviewConfiguration = ({ deviceData, diskSelection, language, localizationData, requests, idPrefix, setIsFormValid, storageScenarioId, accounts }) => {
+const ChangeHostname = ({ initHostname }) => {
+    const [currentHostname, setCurrentHostname] = useState("");
+    const [error, setError] = useState([]);
+    const [isModalOpen, setIsModalOpen] = React.useState(false);
```
KKoukiou, line 68: Let's be consistent on how we use hooks. s/React.useState/useState, like two lines above.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+
+    const validateHostname = (value) => {
+        const error = [];
+        if (value.length > 64) { error.push(_("Real host name must be 64 characters or less")) }
+        if (value.match(/[.a-z0-9-]*/)[0] !== value || value.indexOf("..") !== -1) { error.push(_("Real host name can only contain lower-case characters, digits, dashes, and periods (with populated subdomains)")) }
```
KKoukiou, line 73: Why double dot?

KKoukiou, line 73: The match can return also null, if not matches are found. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value

In this case this would crash.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
 };
 
-export const ReviewConfiguration = ({ deviceData, diskSelection, language, localizationData, requests, idPrefix, setIsFormValid, storageScenarioId, accounts }) => {
+const ChangeHostname = ({ initHostname }) => {
+    const [currentHostname, setCurrentHostname] = useState("");
+    const [error, setError] = useState([]);
+    const [isModalOpen, setIsModalOpen] = React.useState(false);
+
+    const validateHostname = (value) => {
+        const error = [];
```
KKoukiou, line 71: Can you please rename to validationError? error variable is defined already, so we rely on the scoping for this, which is a bit confusing.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+                  </>
+              }
+              id="system_information_change_hostname"
+              isOpen={isModalOpen}
+              onClose={handleModalToggle}
+              position="top"
+              title={_("Change host name")}
+              variant="small"
+            >
+                <Form isHorizontal onSubmit={() => onSubmit}>
```
KKoukiou, line 122: onSubmit expects a function. You can either use onSubmit={onSubmit} or use onSubmit={() => onSubmit()} - notice () after function name.

Please use the former for consistency.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+                    <FormGroup fieldId="review-handle-hostname-hostname" label={_("Host name")}>
+                        <TextInput id="review-handle-hostname-hostname" placeholder={initHostname === "" ? "Default" : initHostname} onChange={(_event, value) => onHostnameChanged(value)} validated={error.length ? "error" : "default"} />
```
KKoukiou, line 124: Default, is not a host name, place holders should use 'examples'. Feel free to drop the placeholder, or communicate with @garrett for an appropriate default value.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+        validateHostname(value);
+        setCurrentHostname(value);
+    };
+    const onSubmit = (event) => {
+        setHostname({ hostname: currentHostname })
+                .then(() => {
+                    handleModalToggle();
+                })
+                .catch(ex => {
+                    console.info({ ex });
```
KKoukiou, line 88: The console.info looks wrong here.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+                });
+
+        if (event) { event.preventDefault() }
+        return false;
+    };
+
+    const handleModalToggle = _event => {
```
KKoukiou, line 95: _event is not used. No need to name it.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -58,8 +62,86 @@ const ReviewDescriptionList = ({ children }) => {
...
+                  <>
+                      <Button variant="primary" isDisabled={disabled} onClick={onSubmit}>{_("Change")}</Button>
+                      <Button variant="link" onClick={handleModalToggle}>{_("Cancel")}</Button>
+                  </>
+              }
+              id="system_information_change_hostname"
+              isOpen={isModalOpen}
+              onClose={handleModalToggle}
+              position="top"
+              title={_("Change host name")}
```
KKoukiou, line 119: Please ask in the channel about anaconda's host name vs hostname convention.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -75,8 +157,30 @@ export const ReviewConfiguration = ({ deviceData, diskSelection, language, local
         setIsFormValid(true);
     }, [setIsFormValid]);
 
+    useEffect(() => {
+        console.log(hostname);
```
KKoukiou, line 161: console.log needs to be removed.

KKoukiou (changes requested):
Some review comments but most importantly this needs communication with @garrett about it's positioning in the review page. Me and @jkonecny12 both feel that's not the correct place.

src/components/review/ReviewConfiguration.jsx
adamkankovsky: It should report an empty hostname, but it's not valid

KKoukiou: But here you match double dots no?

src/components/review/ReviewConfiguration.jsx
garrett: Correct, it shouldn't use a placeholder; it should prefill the existing already-generated hostname, and someone can then edit it or replace it completely with something else.

src/components/review/ReviewConfiguration.jsx
garrett: We had a discussion about this in Slack a while back; I don't remember what the conclusion was... except that we should be consistent. But, then, consistent with what?

So, to capture what I remember of the discussion here....

Let's look at what we ship:

- GNOME uses "Device name".
- KDE... ????? (I can't find it anywhere, oddly. I've looked all over the place.)
- Linux system uses `/etc/hostname` and the `hostname` command. These are from the 1970s.
- Cockpit uses "Hostname" (one word), but this is a reflection of the above UNIX references.

Let's look at what everyone else is doing:

- Windows uses "Device name" (Windows 11); they previously used "Computer name" (Windows 10 and prior)
- macOS uses "Computer name"
- Android uses "Device name"
- iPhone uses "Name" (under General > About), but the phone is referred to "Device" in several other places, and when the name is being referred to in the generic sense, "Device name" is used.

As the most recent versions of GNOME, Windows, Android, and iOS all use "Device name" and also because it's a non-jargon name, I'd ideally prefer if we used "Device name". "Computer" name is similar, but more specific... as Fedora is used on tablets and [gaming consoles](https://bazzite.gg/) as well (which are definitely devices, but not considered "computers"). So, "Device" is more common and inclusive and the better of the two.

However, for historical reasons, "Hostname" is used by UNIX and Linux (which is why it's been referred to as "Hostname" in Cockpit too). It's not "Host name" (two words); it's "Hostname" (one word). But this is computer jargon, and interfaces should generally avoid jargon, except where it's extremely commonplace or unavoidable.

So the question then would be: Should we be more consistent with modern computing in general — using "Device name" — or be consistent with jargon from the 1970s :wink: and go with "Hostname"?

So, saying all that, **in summary**; we have two options: "Device name" (ideally) or "Hostname", _but not "Host name"._

We also talked about where this could be changed. Everything other than the review page would require a new screen. The review screen could be inline editing (which is in PatternFly, but usually not in Anaconda or Cockpit) or a modal (as it does in this PR). We did talk about it possibly also being in a networking page (when we have one, and where we'd have on — like on a server install) or even something like a username page.

https://redhat-internal.slack.com/archives/C04TE3FPD32/p1709802956865689?thread_ts=1709717365.590469&cid=C04TE3FPD32

garrett:
So I quickly tacked on the device name to the overview mockup and got this:

Wizard, review, list(s) (2)

But we might want to reorganize the review screen to have grouped sections, so that configuration is one section and storage is another. I played around with the grouping and order a few times, and I think this might be the most logical?

Wizard, review, list(s) (3)

Justification:

- First is the general section, which is a review of what someone's about to install
- Second section is the configuration, which is in order of what they set. (Language was set first, then account name, and the device name was not manually set but autogenerated so it comes last).
- Then the storage related part, where it's the installation type and the actual storage that will be used. (In the mockup, it's rather complex storage compared to what most people would see, as there are two disks with several partitions.) 

Additionally, there are way more "edit"s than what we are planning to support in the near-term, so it'd really look more like this, for now at least:

image


What do you think?

garrett:
Why isn't there a hostname here? The design completely falls apart if the hostname isn't already predefined. It's also supposed to be "edit" on the review screen, as there should be a hostname — not set. (Setting happens in the modal only, and it's optional.)

image

adamkankovsky:
> Why isn't there a hostname here? The design completely falls apart if the hostname isn't already predefined. It's also supposed to be "edit" on the review screen, as there should be a hostname — not set. (Setting happens in the modal only, and it's optional.)
> 
> image

Here is the problem, because it is still an uninstalled system, so the hostname is not set yet and I understood that if it is not set, there will be a set and if there is, there will be an edit. At the moment there is no mechanism in Anaconda to generate the hostname before the installation.

garrett:
In other words: We should not set the hostname in the review screen. Changing it is OK if it's already set (which it really should be — it needs a default without manually setting it), but it is definitely not OK to have to set it in the review screen.

test/check-review
```diff
@@ -32,7 +32,11 @@ class TestReview(anacondalib.VirtInstallMachineCase):
         m = self.machine
         i = Installer(b, m)
         s = Storage(b, m)
-        r = Review(b)
+        r = Review(b, m)
+
+        # Set default hostname for check
```
KKoukiou, line 37: What happens if you dont set that? Does the backend have some other value autoset?

adamkankovsky: Backend default value is empty value. So hostname will be empty.

garrett:
So we need _something_ there as a hostname. It cannot be black and we cannot force someone to set something for the first time in the review step.

Since we have a transient mode, as long as it's fine as a default (and someone wouldn't _need_ to change it), then we could, perhaps, we can do this as a compromise?

Wizard, review, list(s), device name transient

Set device name modal dialog

KKoukiou (commented):
Please limit this use case to the server in the code (see boot_iso variable) and incorporate the transient hostname suggestion so we can merge this soon.

adamkankovsky:
Updated to new design

rvykydal:
I'd hate to slow down this PR but I'd consider using "network information" instead of "DHCP" as if I understand it correctly it can be set also just from DNS lookup, independently of DHCP (https://www.freedesktop.org/wiki/Software/systemd/hostnamed/ )?

test/check-review
```diff
@@ -95,6 +102,28 @@ class TestReview(anacondalib.VirtInstallMachineCase):
...
+        i.open()
+
+        self.addCleanup(lambda: dbus_reset_users(self.machine))
+        i.reach(i.steps.REVIEW)
+
+        # change hostname in modal
+        b.click("button:contains('edit')")
+        b.set_input_text("#review-handle-hostname-hostname", "Test false", append=False, value_check=False)
+        b.set_input_text("#review-handle-hostname-hostname", "test123", append=False, value_check=True)
+        b.click("button:contains('Save')")
```
rvykydal, line 120: We may also want a case to check for "Change" button label, not sure.

rvykydal (dismissed):
Looks good to me with minor issues I would not block on, but I think we should allow to reset to transient.

adamkankovsky:
> If I change a value from empty/transient to something, then I am not able to change it back to empty/transient. I think we should allow this (and add to tests if so).

Unfortunately there is a bit of a problem here, because Anaconda doesn't allow you to set an empty hostname, so it won't be possible to fix this without modifying the backend.

rvykydal:
> I'd hate to slow down this PR but I'd consider using "network information" instead of "DHCP" as if I understand it correctly it can be set also just from DNS lookup, independently of DHCP (https://www.freedesktop.org/wiki/Software/systemd/hostnamed/ )?
> 
> Other than that I am fine with the design for this PR, thank you.

@garrett what do you think?

rvykydal:
> > If I change a value from empty/transient to something, then I am not able to change it back to empty/transient. I think we should allow this (and add to tests if so).
> 
> Unfortunately there is a bit of a problem here, because Anaconda doesn't allow you to set an empty hostname, so it won't be possible to fix this without modifying the backend.

Hm, I've checked with rawhide that it is possible with the Gtk UI: set it in the entry of Network Spoke to a value, empty string, and then a value again in the backend (setting happens on leaving the Network spoke). Also setting empty hostname is possible. There will be empty /etc/hostname on target system.

adamkankovsky:
> > > If I change a value from empty/transient to something, then I am not able to change it back to empty/transient. I think we should allow this (and add to tests if so).
> > 
> > 
> > Unfortunately there is a bit of a problem here, because Anaconda doesn't allow you to set an empty hostname, so it won't be possible to fix this without modifying the backend.
> 
> Hm, I've checked with rawhide that it is possible with the Gtk UI: set it in the entry of Network Spoke to a value, empty string, and then a value again - the value will propagate in the backend on leaving the Network spoke. Also setting empty hostname is possible. There will be empty /etc/hostname on target system.

@rvykydal  So I finally modified the regex to allow empty hostnames and added tests to test it.

src/components/review/ReviewConfiguration.jsx
```diff
@@ -79,78 +210,130 @@ const ReviewConfiguration = ({ idPrefix, setIsFormValid }) => {
...
+            </FlexItem>
+        </Flex>
     );
 };
 
-export const ReviewConfigurationConfirmModal = ({ idPrefix, onNext, setNextWaitsConfirmation }) => {
-    const { buttonLabel, buttonVariant, dialogTitleIconVariant, dialogWarning, dialogWarningTitle } = useScenario();
-
+export const ReviewConfigurationConfirmModal = ({
+    idPrefix,
```
KKoukiou, line 326: these changes are unrelated and add noise to the commit. Please avoid code-reorganization in feature commits.

KKoukiou (commented):
I split out the hostname related code to seperate component and file for readability purposes. Also the groupping of the description elements should be seperate commit for the same purpose.

KKoukiou (approved):

PR #179, opened 2024-02-16
storage: fix modal backdrops from modals from cockpit-storage to extend to full page

KKoukiou:
Resolves: rhbz#2264015

* [x] Blocked on: https://github.com/cockpit-project/cockpit/pull/20046
* [x] CSS is still  not working

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -43,6 +43,28 @@ import { initialState, reducer, useReducerWithThunk } from "../reducer.js";
 const _ = cockpit.gettext;
 const N_ = cockpit.noop;
 
+const MaybeBackdrop = ({ children }) => {
+    const [hasDialogOpen, setHasDialogOpen] = useState(false);
+
+    useEffect(() => {
+        const handleStorageEvent = (event) => {
+            if (event.key === "cockpit_has_modal") {
+                setHasDialogOpen(event.newValue === "true");
+            }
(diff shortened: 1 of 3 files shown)


garrett:
I figured out a solution and directly inherit from PatternFly for the backdrop so it matches 100% @ https://github.com/rhinstaller/anaconda-webui/pull/179/commits/3844f23e23b121350fb2bb1bbafd73f4a10d22d1.

Please test, as you know my machine is wacky. :wink:

src/components/app.jsx
```diff
@@ -49,6 +49,24 @@ import { useReducerWithThunk, reducer, initialState } from "../reducer.js";
 const _ = cockpit.gettext;
 const N_ = cockpit.noop;
 
+const MaybeBackdrop = ({ children }) => {
+    const [hasDialogOpen, setHasDialogOpen] = useState(false);
+
+    useEffect(() => {
+        window.addEventListener("message", event => {
```
jexposit, line 56: I think that you need to remove this listener when the component unmounts, i.e., return a function in this hook that cleans it.

KKoukiou: Good point, removing unused listeners prevents memory leaks. If this unmounts though it really means the the installer finished, so the user rebooted or so, so there is not big deal there.

I will add it still, does not hurt.

PR #181, opened 2024-02-16
Fix Javascript error hanlding

KKoukiou:
The report dialog was always showing 'undefined' when JS exceptions occured.


adamkankovsky (approved):
I've seen this code before xD. Looks good to me.

PR #182, opened 2024-02-16
storage: inform user when oops happens in cockpit-storage

marusak:


KKoukiou (commented):
Please pass the while exception object like this:
```
iframe.contentWindow.addEventListener("error", onCritFail({ context: N_("Storage plugin failed") }));
```

This will automatically pass it. The reason I want that, is that the exception object contains not only the meesage but the while StackTrace in property, stack. Did not test this here with this eventlistener, but normally it does. 

We want the tracktrace, for debugging purposes, as this will be eventually auto-reported to BZ.

Take in mind, since we know this is an error in cockpit-storage, we should change the default component from anaconda-webui to cockpit. Let's do the latter in followup.

marusak:
Updated

KKoukiou (approved):
I would like to have onCritFail handle conditionally also JS exceptions, so we have a unified hanlder. Lets get this in as it's a great bugfix and I will send a followup.

PR #187, opened 2024-02-18
storage: keep all scenarios in the partitioning object mapping

KKoukiou:
This prevents the 'Use configured storage' scenario from disappearing when selecting another scenario. The 'Use configured storage' should disappear when the applied partitioning is not longer the one created from the storage page.


rvykydal (approved):
Looks good to me.

KKoukiou:
Did clean rebase to let tests run again with some test stabiliazation which got just merged.

PR #189, opened 2024-02-18
tests: don't test the order of the storage devices in the review page

KKoukiou:
They seem to be shuffling around sometimes and also it's not important.


KKoukiou:
Merged as part of https://github.com/rhinstaller/anaconda-webui/pull/182 to avoid waiting for the tests again.

PR #193, opened 2024-02-19
WIP: Trying to de-emphasize the option descriptions.

garrett:
I'm trying to make the options less emphasized, as having them with darker text and bolding make them too emphasized, especially when the option is in an inactive state.

This is especially an issue when some other item is selected, such as "Use configured storage". The "Not enough free space" message _really_ jumps out and looks relevant to the situation due to it being:
1. Darker (not using the same description text color)
2. Bolder.
3. Very wordy.

Here's what I'm trying to do (mockup):

Welcome, disk, method_ configured storage replacement (8)

Here's what it currently looks like (outside of this PR):

image

### DIFF
--- a/src/components/storage/InstallationScenario.jsx
+++ b/src/components/storage/InstallationScenario.jsx
@@ -70,9 +70,9 @@ export const checkUseFreeSpace = ({ diskFreeSpace, diskTotalSpace, requiredSize 
     }
     if (diskFreeSpace < requiredSize) {
         availability.available = false;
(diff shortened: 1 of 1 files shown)


garrett:
This was an attempt where I tried to change the code to make it less obvious, but it needed help.

With the upcoming changes with reclaim, part of this is no longer valid. The rest is no longer valid when we adopt the recent mockups (especially the ones with rewording).

So I'm closing this, as we'll solve the problems in other ways. :wink:

PR #194, opened 2024-02-20
Dark mode support

adamkankovsky:


KKoukiou:
@adamkankovsky I was about to merge this, but it's  not conflicting as i merged Radeks PR first which also edited test/reference. Can you re-fetch.?

PR #195, opened 2024-02-20
test: Cockpit now leaves LUKS devices open after creation

mvollmer:
The workaround can be removed.

This needs to go in together with https://github.com/cockpit-project/cockpit/pull/20016.


KKoukiou (approved):

mvollmer:
@KKoukiou, oh, this passes even without https://github.com/cockpit-project/cockpit/pull/20016... Why doesn't it fail? So. can we just merge this?

PR #213, opened 2024-03-07
Enforce some more eslint rules and introduce some minor cleanups

KKoukiou:


rvykydal (approved):
Looks good to me, thank you.

(Regarding the sort-keys rule, in some cases I've got a feeling it prevents expressing um - something? by the order order of keys in the code (like most important keys / information first when reading the code maybe) but it is a very vague feeling and the order gets quite random as the keys are added incrementally anyway, so fine by me.)

KKoukiou:
> Looks good to me, thank you.
> 
> (Regarding the sort-keys rule, in some cases I've got a feeling it prevents expressing um - something? by the order of the keys in the code (like most important keys / information first when reading the code maybe) but it is a very vague feeling and the order gets quite random as the keys are added incrementally anyway, so fine by me.)

I see your point, I dont think we ever put much effort on keeping keys organized this way though, so I also tend to think it's a small readability improvement, to keep the keys at least alphabetically sorted.

PR #220, opened 2024-03-13
Group and sort imports by from * values

KKoukiou:


KKoukiou (commented):
Just for clarification, the imports are grouped and sorted by the from * values.

The code was obviously autofixed, I did not so this manually.

marusak (approved):
Danke! :)

PR #249, opened 2024-04-10
storage: cockpit-integration: do not overwrite request object structure - only extend it

KKoukiou:
Resolves: rhbz#2264419


rvykydal (approved):
Looks good to me.

PR #274, opened 2024-05-10
Move stepNotification to the AnacondaPage component

KKoukiou:
These notifications are page specific are belong better to the page wrapper, rather than the global AnacondaWizard component.

With this change the notifications are now reset / removed when changing pages. This is the desired behavior, as failures should appear only dynamically after clicking 'Next'.


rvykydal (approved):
Looks good to me.

PR #286, opened 2024-05-23
storage: Introduce 'Reclaim space' dialog

KKoukiou:


src/actions/storage-actions.js
```diff
@@ -43,8 +43,9 @@ export const getDevicesAction = () => {
     return async (dispatch) => {
         try {
             const devices = await getDevices();
+            const deviceData = {};
             const mountPoints = await getMountPoints();
-            const devicesData = await Promise.all(devices.map(async (device) => {
+            for (const device of devices) {
```
rvykydal, line 51: Just reading loud, so this will also gather the data serially instead of concurrently, which is OK (perhaps even desirable to be safer) I think.

src/components/storage/Common.jsx
```diff
@@ -128,3 +139,17 @@ export const useMountPointConstraints = () => {
 
     return mountPointConstraints;
 };
+
+export const getNewPartitioning = async ({ currentPartitioning, method = "AUTOMATIC" }) => {
+    await setInitializeLabelsEnabled({ enabled: true });
```
rvykydal, line 144: I am wondering if this should be true only for AUTOMATIC.
In TUI we have this rule: https://github.com/rhinstaller/anaconda/blob/a73fe25bfb521fbf59f24b991a83cd5e9c2cb9d4/pyanaconda/ui/tui/spokes/storage.py#L521
(but in TUI the alternative partitioning is only MANUAL).
In GUI seemingly not but I am not sure:  https://github.com/rhinstaller/anaconda/blob/a73fe25bfb521fbf59f24b991a83cd5e9c2cb9d4/pyanaconda/ui/gui/spokes/storage.py#L226
(maybe it is harmless for other cases)

In the end the method feels more like getNewAutomaticPartitioning at this point, but perhaps it can be used for other types of partitioning in the future.

Anyway, I wouldn't block the PR on clearing up this, I'd need to figure out myself how this works exactly for various use cases.

src/apis/storage_devicetree.js
```diff
@@ -20,14 +20,30 @@ import { StorageClient } from "./storage.js";
...
-};
-const callHandler = (...args) => {
-    return _callClient(StorageClient, OBJECT_PATH, INTERFACE_NAME_HANDLER, ...args);
-};
+export class DeviceTree {
+    constructor (deviceTree = OBJECT_PATH) {
+        if (DeviceTree.instance && DeviceTree.instance.deviceTree === deviceTree) {
+            return DeviceTree.instance;
+        }
+        DeviceTree.instance = this;
```
rvykydal, line 31: I can't figure out what is the role of instance here...

src/components/storage/DiskEncryption.jsx
```diff
@@ -70,11 +70,17 @@ const DiskEncryption = ({
...
-    const [password, setPassword] = useState(request?.passphrase || "");
+    const [confirmPassword, setConfirmPassword] = useState();
+    const [isEncrypted, setIsEncrypted] = useState();
+    const [password, setPassword] = useState();
     const luksPolicy = useContext(RuntimeContext).passwordPolicies.luks;
 
+    useEffect(() => {
+        setIsEncrypted(request?.encrypted);
+        setPassword(request?.passphrase);
+        setConfirmPassword(request?.passphrase);
```
rvykydal, line 81: I guess we are fine with not requiring string here (dropping the `|| ""`) part.

src/apis/storage_devicetree.js
KKoukiou: NOt needed - not sure what I had in mind when I wrote the class, removing. thanks :)

src/components/storage/DiskEncryption.jsx
KKoukiou: I noticed that from the API these are empty strings by default.
https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/common/structures/partitioning.py#L38

so no need for this.
I will split this change is seperate commit for clarify, it's not nice that I hide this in this massive PR.

rvykydal: Thank you, feel free to keep it, fine by me.

rvykydal:
When playing with the PR locally I hit a situation where unchecking the "Reclaim" checkbox moves user automatically to the next page:
https://rvykydal.fedorapeople.org/webui/reclaim_dialog/unchecking_reclaim_results_in_next.webm

test/check-storage
```diff
@@ -266,15 +266,14 @@ class TestStorage(anacondalib.VirtInstallMachineCase, StorageHelpers):
             s.dbus_create_partitioning("AUTOMATIC")
 
         # Go back to the Disk Configuration page and re-enter the review screen.
-        # This should create again a new partitioning object and apply it
-        # no matter how many partitioning objects were created before
+        # This should create a new partitioning object since we always re-create automatic partitiooning:
```
rvykydal, line 269: partitiooning :)

rvykydal (commented):
Looks great to me, have some minor issues or questions, the only thing that prevented me from Approving is the issue with automatic Next I described in a comment:

> When playing with the PR locally I hit a situation where unchecking the "Reclaim" checkbox moves user automatically to the next page:
> https://rvykydal.fedorapeople.org/webui/reclaim_dialog/unchecking_reclaim_results_in_next.webm
> If it is difficult to reproduce on your side, I can play with it more and try to provide more specific reproducer, maybe my local issue or weird use case.

... so waiting for you to check or comment.

rvykydal:
A minor issue that can be definitely fine-tuned later if at all is - I'd add a bit more padding of vda1 and vda2. Just my 2 cents regarding the layout on which I think others may comment more later. I am not focusing on it in this review and definitely not blocking from my side. (Also may be enough of padding given there can be more levels).
padding_vda1_vda2

src/components/storage/DiskEncryption.jsx
KKoukiou: Split

src/actions/storage-actions.js
KKoukiou: Yes, honestly I did not serialize this on the purpose. I was trying to clean up the code and make it a bit more readable. I guess it's ok side effect, we dont have performance issues here.

rvykydal:
Just quickly re-checked, but I guess you'll re-require review when the PR is ready. For now the availability of the scenarios seems to be broken at least for single disk use case.

rvykydal (approved):
Looks good to me.

PR #307, opened 2024-06-18
test: Ignore cockpit-storage frame in pixel test

mvollmer:
This makes it easier to change Cockpit without having to worry about Anaconda tests.

### DIFF
--- a/test/check-storage
+++ b/test/check-storage
@@ -1112,7 +1112,8 @@ class TestStorageCockpitIntegration(anacondalib.VirtInstallMachineCase, StorageC
         b.assert_pixels(
             "#app",
             "storage-editor",
-            ignore=[*anacondalib.pixel_tests_ignore, ".cockpit-storage-integration-requirements-hint"],
+            ignore=[*anacondalib.pixel_tests_ignore, ".cockpit-storage-integration-requirements-hint",
+                    "#cockpit-storage-frame"],
         )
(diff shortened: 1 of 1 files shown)


KKoukiou (commented):
Hmm @mvollmer are you sure about this? We do on the purpose test cockpit storage from within anaconda tests. And we set up the anaconda-webui tests on cockpit repository for this purpose. Do you find the pixel only especially annoying or generally the whole part of cross repo testing?

mvollmer:
> Hmm @mvollmer are you sure about this? 

Not very, tbh... As you say, pixels are only one aspect. If the Cockpit UI changes in larger ways, like moving actions from one menu to another, then the Anaconda tests will need updating as well.

But there is no good way now to handle changes to Cockpit now, right?  The Anaconda tests need Cockpit from rawhide. That means they will very easily fail in Cockpit PRs like here: https://github.com/cockpit-project/cockpit/pull/20611, https://cockpit-logs.us-east-1.linodeobjects.com/pull-20611-19e362fa-20240618-121735-fedora-rawhide-boot-cockpit-pr-20611-rhinstaller-anaconda-webui/log.html

(If Cockpit would be a library and Anaconda would use its stable API, then any change in Cockpit that makes the Anaconda tests fail is a regression. But we don't want to make any stability promising about our UI... hmm.)

PR #309, opened 2024-06-18
Cleanup code around step management

KKoukiou:


adamkankovsky (approved):
It probably actually makes more sense this way.

rvykydal:
Looks good to me.

PR #312, opened 2024-06-19
storage: reset the partitioning when mounting DiskEncryption

KKoukiou:
Otherwise applying the new one is problematic for the 'use-free-space' scenario.


adamkankovsky (approved):
LGTM, thanks.

PR #321, opened 2024-06-26
components: standardize page initialization

KKoukiou:


src/components/AnacondaPage.jsx
```diff
@@ -14,12 +14,20 @@
...
-export const AnacondaPage = ({ children, step, title }) => {
+export const AnacondaPage = ({ children, setIsFormDisabled, step, title, usePageInit }) => {
     const [stepNotification, setStepNotification] = useState();
 
+    // If there is usePageInit custom hook for the page, call it
+    usePageInit?.();
+
+    // Otherwise just set the form to enabled so that the user can interact with it
+    useEffect(() => {
+        if (!usePageInit) setIsFormDisabled(false);
```
martinpitt, line 28: Missing line break

src/components/AnacondaPage.jsx
```diff
@@ -14,12 +14,20 @@
...
+    }, [setIsFormDisabled, usePageInit]);
```
martinpitt, line 29: The `setIsFormDisabled` dependency here is really weird -- that function never changes. It would make more sense to depend on changes to the actual value. But I suppose that's just  React not understanding the bigger picture, so just a formality -- and the old code you removed had it as well. `</thinking-aloud>`.

martinpitt (approved):
Looks sane enough, thanks!

src/components/AnacondaPage.jsx
KKoukiou: It's just formality to silence the linter indeed.

src/components/AnacondaPage.jsx
KKoukiou: Fixing this @ https://github.com/rhinstaller/anaconda-webui/pull/323 so that I wait one time for the tests.

PR #323, opened 2024-06-26
Generic improvements around page initialization and definitions

KKoukiou:
Laying on top of https://github.com/rhinstaller/anaconda-webui/pull/321

### DIFF
--- a/src/components/AnacondaPage.jsx
+++ b/src/components/AnacondaPage.jsx
@@ -14,11 +14,21 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
-import React, { cloneElement, useState } from "react";
+import React, { cloneElement, useEffect, useState } from "react";
 import { Alert, Stack, Title } from "@patternfly/react-core";
 
-export const AnacondaPage = ({ children, step, title }) => {
+export const AnacondaPage = ({ children, setIsFormDisabled, step, title, usePageInit }) => {
     const [stepNotification, setStepNotification] = useState();
+
+    // If there is usePageInit custom hook for the page, call it
+    usePageInit?.();
+
+    // Otherwise just set the form to enabled so that the user can interact with it
+    useEffect(() => {
+        if (!usePageInit) {
+            setIsFormDisabled(false);
+        }
+    }, [setIsFormDisabled, usePageInit]);
 
     return (
         <Stack hasGutter>
--- a/src/components/AnacondaPage.jsx
+++ b/src/components/AnacondaPage.jsx
@@ -14,11 +14,13 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
-import React, { cloneElement, useEffect, useState } from "react";
+import React, { cloneElement, useEffect, useRef, useState } from "react";
 import { Alert, Stack, Title } from "@patternfly/react-core";
 
-export const AnacondaPage = ({ children, setIsFormDisabled, step, title, usePageInit }) => {
+export const AnacondaPage = ({ children, isFormDisabled, setIsFormDisabled, step, title, usePageInit }) => {
     const [stepNotification, setStepNotification] = useState();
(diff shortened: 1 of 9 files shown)


src/components/AnacondaPage.jsx
```diff
@@ -37,13 +37,15 @@ export const AnacondaPage = ({ children, isFormDisabled, setIsFormDisabled, step
         }
     }, [isFormDisabled]);
 
+    const titleIsEmpty = title?.type?.() === null || !title;
```
martinpitt, line 40: How is that different from just `!title`? `!null === true`

martinpitt (commented):
This will need rebasing after #321, but one question in the meantime.

src/components/AnacondaPage.jsx
KKoukiou: title can be a null React.component as of https://github.com/rhinstaller/anaconda-webui/pull/323/commits/d7b86c791d9d84ebbd60cd5038ec01345baf25a1#diff-7768fc0c14b2131f2f8fe93ac2e875fb66357a0766198d965f8f5f35a308167bR202

But actually I will adjust and allow title to be only string. Reposting.

martinpitt: Yes, I'm asking why you introduced the `title?.type?.() === null` check. How is that different from `!title`?

KKoukiou: Because a React.component that returns null is not === null, it's still a React.component so some Object type, so !title will be truthy :)
Moved the null check out, see repush.

src/components/AnacondaPage.jsx
```diff
@@ -30,6 +32,17 @@ export const AnacondaPage = ({ children, setIsFormDisabled, step, title, usePage
...
 
+    useEffect(() => {
+        if (!isFormDisabled && !showPageRef.current) {
+            showPageRef.current = true;
+            setShowPage(true);
+        }
+    }, [isFormDisabled]);
+
+    if (!showPage) {
+        return null;
```
martinpitt, line 43: Could it happen that initialization takes very long (like the Storage page)? Staring at a blank page isn't very nice either, perhaps this should be rendered to an empty state after 100ms timeout or so?

src/components/localization/InstallationLanguage.jsx
```diff
@@ -329,15 +329,17 @@ const InstallationLanguage = ({ idPrefix, setIsFormValid, setStepNotification })
...
-        component: InstallationLanguage,
-        id: "installation-language",
-        isHidden: !isBootIso,
-        label: _("Welcome"),
-        title: cockpit.format(_("Welcome to $0"), osRelease.NAME),
-    });
+    return cockpit.format(_("Welcome to $0"), osRelease.NAME);
 };
+
+export class Page {
```
martinpitt, line 337: The "convert each page definition to a class" commit could use some justification, i.e. the commit message could say "why?". That's not obvious.

src/components/AnacondaPage.jsx
martinpitt: Hmm, but wouldn't that be more like `title?.().type === null`? (That still doesn't cover calling a render function or something, just a functional component). Also, that would be `title?() === null`. 

To me this is absolutely incomprehensible, and it should get a comment what exactly happens here. Anyway, it's moot now as you dropped it.

martinpitt (approved):
Thanks! No big objections any more, go ahead if this blocks you.

src/components/AnacondaPage.jsx
KKoukiou: It's not noticeable by the user at this stage, but it can lead into race conditions inside pages, if we dont guard it. Happy to send a follow up to show the empty state.

src/components/localization/InstallationLanguage.jsx
KKoukiou: It's an architectural decision that I arbitrarily met to improve readability. I am open to discuss this. Generally I want that each Wizard page has an standardized exported 'something', ie class, object whatever, so that the Wizard can be agnostic of the pages is contains. 

This opens the road for supporting in the future add-ons.

PR #327, opened 2024-06-27
Fix wrong background when dialog in the cockpit storage modal view is…

KKoukiou:
After we ported the cockpit storage view to a modal, we forgot to adjust the CSS that handles the backdrop.

Screen Shot 2024-06-27 at 12 26 48

### DIFF
--- a/src/components/Common.jsx
+++ b/src/components/Common.jsx
@@ -14,7 +14,7 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with This program; If not, see <http://www.gnu.org/licenses/>.
  */
-import React, { createContext, useEffect, useState } from "react";
+import React, { createContext } from "react";
 import { Popover, PopoverPosition } from "@patternfly/react-core";
 import { HelpIcon } from "@patternfly/react-icons";
 
@@ -79,38 +79,14 @@ const SystemInfoContextWrapper = ({ children, conf, osRelease }) => {
     );
 };
 
-const MaybeBackdrop = ({ children }) => {
-    const [hasDialogOpen, setHasDialogOpen] = useState(false);
-
-    useEffect(() => {
-        const handleStorageEvent = (event) => {
-            if (event.key === "cockpit_has_modal") {
-                setHasDialogOpen(event.newValue === "true");
-            }
-        };
-
-        window.addEventListener("storage", handleStorageEvent);
-
-        return () => window.removeEventListener("storage", handleStorageEvent);
-    }, []);
-
-    return (
-        <div className={hasDialogOpen ? "cockpit-has-modal" : ""}>
-            {children}
-        </div>
-    );
-};
-
 export const MainContextWrapper = ({ address, children, conf, osRelease, state }) => {
     return (
         <ModuleContextWrapper state={state}>
             <SystemInfoContextWrapper osRelease={osRelease} conf={conf}>
                 <WithDialogs>
-                    <MaybeBackdrop>
-                        <AddressContext.Provider value={address}>
-                            {children}
-                        </AddressContext.Provider>
-                    </MaybeBackdrop>
+                    <AddressContext.Provider value={address}>
+                        {children}
+                    </AddressContext.Provider>
                 </WithDialogs>
             </SystemInfoContextWrapper>
         </ModuleContextWrapper>
--- a/src/components/storage/CockpitStorageIntegration.scss
+++ b/src/components/storage/CockpitStorageIntegration.scss
@@ -1,3 +1,6 @@
+@import "global-variables";
+@import "@patternfly/patternfly/components/Backdrop/backdrop.scss";
+
 .cockpit-storage-integration-modal-page-section {
     width: 100%;
     height: 100%;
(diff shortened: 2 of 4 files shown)


src/components/Common.jsx
```diff
@@ -79,7 +79,7 @@ const SystemInfoContextWrapper = ({ children, conf, osRelease }) => {
     );
 };
 
-const MaybeBackdrop = ({ children }) => {
+export const useMaybeBackdrop = () => {
     const [hasDialogOpen, setHasDialogOpen] = useState(false);
 
     useEffect(() => {
```
martinpitt, line 85: I find this super weird. On the outside this is now a simple query function which returns a class name. But each time this gets called by the storage component (i.e. on each render), this will set up a new useEffects(). This doesn't feel healthy, and it feels like that effect should be pulled out of this function.

I already hate useEffects(), but this feels too inflationary.

martinpitt (changes requested):
Sorry..

src/components/Common.jsx
KKoukiou: useMaybeBackdrop is a custom hook (https://react.dev/reference/rules/rules-of-hooks) with no dependencies, so it will run its useEffect only once, no matter how many re-renders the parent does.

I think actually this usage of hooks as I do it here, is the idea of custom hooks.

Btw I verified my thoughts by putting a console log inside the useMaybeBackdrop useEffect, and it indeed gets called only once.

Let's discuss this over chat if  you still disagree.

KKoukiou: In any case, because we dont need to wrap the whole app now with that class, I can extract the logic into the CockpitStorageIntergration component itself, and eliminate the hook if you prefer.

KKoukiou: @martinpitt btw this is the same exact concept as https://github.com/cockpit-project/cockpit/blob/main/pkg/lib/hooks.ts#L65 for example.

martinpitt: OK! effects keep confusing me :cry:

martinpitt (approved):

KKoukiou (commented):
I still want review from @garrett 
I see it works (see pixel test) but I dont understand this CSS.

garrett (approved):
I tested this locally and it looks correct. Thanks!

src/components/storage/CockpitStorageIntegration.scss
```diff
@@ -57,3 +60,25 @@ ul.cockpit-storage-integration-requirements-hint-list {
         color: unset;
     }
 }
+
+// Simulate Backdrop behavior for modals in the iframe
+%iframe-zindex {
+  position: relative;
+  z-index: 10;
+}
```
garrett, line 68: This is SCSS that says "here's some reusable SCSS code; assign it a name %iframe-zindex to be included elsewhere"

src/components/storage/CockpitStorageIntegration.scss
```diff
@@ -57,3 +60,25 @@ ul.cockpit-storage-integration-requirements-hint-list {
...
+
+iframe {
+  @extend %iframe-zindex;
```
garrett, line 71: `@extend` means to use the (S)CSS with a specific name

src/components/storage/CockpitStorageIntegration.scss
```diff
@@ -57,3 +60,25 @@ ul.cockpit-storage-integration-requirements-hint-list {
...
+}
+
+// Simulate a PF background scrim
+.cockpit-has-modal {
+  @extend %iframe-zindex;
```
garrett, line 76: We're reusing that same SCSS here.

After the two `@extend`s are used, we basically get this CSS:

```css
iframe, .cockpit-has-modal {
  position: relative;
  z-index: 10;
}
```

src/components/storage/CockpitStorageIntegration.scss
```diff
@@ -57,3 +60,25 @@ ul.cockpit-storage-integration-requirements-hint-list {
...
+
+  &::after {
+    @extend .pf-v5-c-backdrop;
```
garrett, line 79: This `@extend` is different — it's using a PF class from CSS. (I think we're using precompiled CSS for PF? Either way, this is from some (S)CSS from PF which is known by SASS when this is getting compiled).

This looks like 

```scss
.pf-v5-c-backdrop, .cockpit-has-modal::after {
  /* PF specific code here */
}
```

So this means we're using exactly what PatternFly uses for the scrim, without having to duplicate the CSS and have drift if/when they change it. It will always be the exact same CSS as PF.

src/components/storage/CockpitStorageIntegration.scss
```diff
@@ -57,3 +60,25 @@ ul.cockpit-storage-integration-requirements-hint-list {
...
+    --pf-v5-c-backdrop--Position: absolute;
+    content: "";
+    z-index: 9;
```
garrett, line 82: This block changes PF's position variable to always be absolute (which is what we want — it needs to have position to have a z-index take effect and it should be positioned absolutely (relative to the parent with position, as usual)). It then changes the content to blank, and adjusts the z-index to 9 to be below the iframe (which is at 10, as specified above).

src/components/storage/CockpitStorageIntegration.scss
garrett: Here's some docs on `@extend`, with the first showing a % block as an extended block:

https://sass-lang.com/guide/#extend-inheritance
https://sass-lang.com/documentation/at-rules/extend/

garrett:
I added in some comments to explain the CSS. We might want to consider adding that into the code as comments. (I didn't add it previously as I thought it was self-explanitory, but I know CSS and SCSS pretty well, and I guess it isn't that straightforward otherwise.)

PR #329, opened 2024-06-28
test: introduce test case for installation against a disk with an existing Linux OS

KKoukiou:


test/check-existing-system
```diff
@@ -0,0 +1,51 @@
...
+from installer import Installer
+from review import Review
+from testlib import test_main  # pylint: disable=import-error
+
+WEBUI_TEST_DIR = os.path.dirname(__file__)
+ROOT_DIR = os.path.dirname(WEBUI_TEST_DIR)
+BOTS_DIR = f'{ROOT_DIR}/bots'
+
+class TestExistingSystemFedora(anacondalib.VirtInstallMachineCase):
+    disk_image = os.path.join(BOTS_DIR, "./images/fedora-rawhide")
```
KKoukiou, line 30: @martinpitt i think I would need some discussion / help on how to convert this to non-destructive.

Let me shortly explain: right now all anaconda tests are running on top of empty qcow2 images and booting the TEST_OS=fedora-rawhide-boot ISO.

In this PR I am however introducing new test cases that run on top already installed images. I started here on top of our rawhide image as a backing_store, but I plan to extend against other images as well, including windows if possible.

My problem is that this test class can't be made easily non-destructive, however all tests inside this class, would nicely share one VM if the non-destructive logic was limited to only take tests inside a specific class.

So I would need here something like @nondestructive_inside_class or similar.

Does this make sense? Do you have any thoughts on this?

martinpitt: I'm afraid I don't understand this yet. Let's talk in gmeet today?

PR #334, opened 2024-07-01
 test: introduce test case for installation against a disk with an existing Linux OS

KKoukiou:


KKoukiou (commented):
The disk we attach with virt-xml in each test setup needs to be somehow marked as the one we boot from after reboot. Not sure how to do that. @martinpitt do you know?

test/anacondalib.py
```diff
@@ -14,11 +14,15 @@
...
 
 import os
+import subprocess
 import sys
-from subprocess import CalledProcessError
+import tempfile
 
 # import Cockpit's machinery for test VMs and its browser test API
 TEST_DIR = os.path.dirname(__file__)
+WEBUI_TEST_DIR = os.path.dirname(__file__)
```
martinpitt, line 23: That's the same as the existing `TEST_DIR`, why have a different one?

test/anacondalib.py
```diff
@@ -50,6 +56,17 @@ def setUp(self):
 
         super().setUp()
 
+        # Add installation target disk
+        backing_file = None if not self.disk_image else os.path.join(BOTS_DIR, f"./images/{self.disk_image}")
+        self.add_disk(15, backing_file)
```
martinpitt, line 61: that's 15 GB? (passed to `--disk`). What does that mean with an overlay? My gut feeling is that the size just gets ignored then?

test/anacondalib.py
```diff
@@ -50,6 +56,17 @@ def setUp(self):
...
+        # Select the disk as boot device
+        self._execute(f"virt-xml -c qemu:///session {self.machine.label} --edit --boot hd")
+
+        m = self.machine
+        b = self.browser
+        s = Storage(b, m)
+        s.dbus_scan_devices()
```
martinpitt, line 68: You can write this in one line without the variables, but the more interesting question is: Is anaconda already running at this point? I suppose yes, otherwise this shouldn't be necessary.

test/anacondalib.py
```diff
@@ -60,6 +77,30 @@ def setUp(self):
             self.allow_browser_errors(".*client closed.*")
             self.allow_browser_errors(".*Server has closed the connection.*")
 
+    def _execute(self, cmd):
+        return subprocess.check_call(cmd, stderr=subprocess.STDOUT, shell=True)
```
martinpitt, line 81: I recommend to *not* redirect stderr. It's much more useful to see the error messages in the test log if this fails, than them getting hidden (as nothing happens with the return value).

Also, all calls to _execute() are a single program with arguments, no actual shell constructs. So dropping `shell` and passing an array is cleaner. It's ok for a test, just something to consider.

test/anacondalib.py
```diff
@@ -60,6 +77,30 @@ def setUp(self):
...
+        self._execute(f"virt-xml -c qemu:///session {self.machine.label} --update --add-device --disk {image},format=qcow2,size={size}")
+
+        if self.is_nondestructive():
+            self.addCleanup(self.rem_disk, image)
+
+        return image
+
+    def rem_disk(self, disk):
+        self._execute(f"virt-xml -c qemu:///session {self.machine.label} --update --remove-device --disk {disk}")
+        self._execute(f"rm -f {disk}")
```
martinpitt, line 94: `os.remove(disk)`

test/anacondalib.py
```diff
@@ -60,6 +77,30 @@ def setUp(self):
...
+
+    def _create_disk_image(self, size, image_path=None, quiet=False, backing_file=None):
+        if not image_path:
+            _, image_path = tempfile.mkstemp(suffix='.qcow2', prefix=f"disk-anaconda-{self.machine.label}")
```
martinpitt, line 98: The directory defaults to /tmp, you need to be a bit careful as that's a tmpfs and easily overflows (or at least unnecessarily eats lots of RAM). This is the kind of "large file" that belongs into `dir="/var/tmp"`.

martinpitt (changes requested):
Excellent! This is a much nicer approach. Just some minor local issues.

As you have two nond tests now, does that mean that restarting anaconda in between these works ok?

test/anacondalib.py
KKoukiou: Anaconda is running and the Web UI is ready to be interacted with. 
This is waited for in the VM spawning script https://github.com/rhinstaller/anaconda-webui/blob/main/test/machine_install.py#L170

test/anacondalib.py
KKoukiou: You re right, adjusted.

test/anacondalib.py
KKoukiou: Removed.

test/anacondalib.py
KKoukiou: Replaced as suggested.

test/anacondalib.py
KKoukiou: :+1:

test/anacondalib.py
KKoukiou: Fixed.

KKoukiou:
> Excellent! This is a much nicer approach. Just some minor local issues.
> 
> As you have two nond tests now, does that mean that restarting anaconda in between these works ok?

We have much more than two nondestructive tests. Actually all of them apart from 3-4 are non-destructive.
We don't restart anaconda between test runs, we just never start the installation, so there is not destructive actions happening.
We reset through the APIs all user selections, such as language, account creation and storage planning.

test/anacondalib.py
martinpitt: It was just a question, it looks a bit strange now -- but I'll comment inline.

martinpitt: Actually it's fine, it just looked confusing in the interdiff.

test/anacondalib.py
```diff
@@ -60,6 +76,31 @@ def setUp(self):
...
+        self._execute(f"virt-xml -c qemu:///session {self.machine.label} --update --remove-device --disk {disk}")
+        os.remove(disk)
+
+    def _create_disk_image(self, size, image_path=None, quiet=False, backing_file=None):
+        if not image_path:
+            _, image_path = tempfile.mkstemp(suffix='.qcow2', prefix=f"disk-anaconda-{self.machine.label}")
+        quiet = "-q" if quiet else ""
+        backing_file = f"-o backing_file={backing_file},backing_fmt=qcow2" if backing_file else ""
+        size = f"{size}G" if size else ""
+        self._execute(f"qemu-img create -f qcow2 {backing_file} {quiet} {image_path} {size}")
```
martinpitt: Ah, you are playing a little dirty shell trick here with an empty size ... :grin:

test/anacondalib.py
```diff
@@ -76,29 +79,34 @@ def setUp(self):
...
-        quiet = "-q" if quiet else ""
-        backing_file = f"-o backing_file={backing_file},backing_fmt=qcow2" if backing_file else ""
-        size = f"{size}G" if size else ""
-        self._execute(f"qemu-img create -f qcow2 {backing_file} {quiet} {image_path} {size}")
+        subprocess.check_call(
+            ["qemu-img", "create", "-f", "qcow2"] +
+            (["-o", f"backing_file={backing_file},backing_fmt=qcow2"] if backing_file else []) +
+            [image_path] +
+            ([f"{size}G"] if size else [])
+        )
```
martinpitt, line 109: Ah, and that cleans it up again :grin: This works. Just note that there is no need to do the multiple [] + [] -- you can just break lines inside of an array:
```py
check_call([
   "aaa", "bbb",
   "--cccc=foo",
])
```

and just add the last bit. If you want to avoid that as well, you can do

    [..., image_path, *([f"{size}G"] if size else []) ]

but that's completely optional.

KKoukiou: https://github.com/rhinstaller/anaconda-webui/pull/335

PR #339, opened 2024-07-04
review: show existing systems affected by deleted partitions

KKoukiou:
127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html (1)

127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html (1)

### DIFF
--- a/src/actions/storage-actions.js
+++ b/src/actions/storage-actions.js
@@ -23,6 +23,7 @@ import {
     getDevices,
     getDiskFreeSpace,
     getDiskTotalSpace,
+    getExistingSystems,
     getFormatData,
     getMountPoints,
 } from "../apis/storage_devicetree.js";
@@ -53,6 +54,7 @@ export const getDevicesAction = () => {
             const devices = await getDevices();
             const deviceData = {};
             const mountPoints = await getMountPoints();
+            const existingSystems = await getExistingSystems();
             for (const device of devices) {
                 try {
                     const devData = await getDeviceData({ disk: device });
@@ -87,6 +89,7 @@ export const getDevicesAction = () => {
                     actions,
                     deviceNames: devices,
                     devices: deviceData,
+                    existingSystems,
                     mountPoints,
                 },
                 type: "GET_DEVICES_DATA"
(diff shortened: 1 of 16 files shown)


garrett:
In the case of deleting a partition that can break an installed OS that will still persist in some residual form (that is, with some partitions associated with it that will still be around and not used by the installation), it should probably look more like this, and be a warning string (as it would actually break functionality):

- Deletion of certain partitions may prevent Fedora 41 from booting: sda3 (var), sda4 (usr)

(This would not show up if you're removing the main partition, such as what is mounted as /, as that would be removing the OS itself.)

When operating systems are to be removed (the main partition, at least), it should be like this:

- Fedora Linux 41 will be deleted
- Windows 11 will be deleted
- macOS 14 Sonoma will be deleted

When recovery partitions are to be deleted, but the OS persists, you'd see something like:

- Windows 11 recovery partition will be deleted. Removing this partion may limit your ability to repair or reinstall Windows.
- macOS 14 Sonoma recovery partition will be deleted. Removing this partion may limit your ability to repair or reinstall macOS.

We could either have the first sentence shown as a phrase (without a period) and with an (i) for the second sentence to explain what this means, or we could have two sentences next to each other like the above examples.

KKoukiou:
@garrett Let's please not introduce logic for what might be destructive partition removal or not. Let's just detect if the whole OS was whipped out or part of it and show the relevant message.

garrett:
> Let's please not introduce logic for what might be destructive partition removal or not. Let's just detect if the whole OS was whipped out or part of it and show the relevant message.

Sure, I agree with this and only included it as you had it in your screenshot at the top of this PR. (I'm referring to the first screenshot, which has the very verbose line saying "You have selected to delete certain partitions associated with your existing operating system Fedora Linux 41 for x86_64. This action may result in data loss and could render your current system unbootable. The affected partitions are: root, home, var")

KKoukiou:
@garrett sure - I think it's important to differentiate between full and partial removal. But I just say that I dont want to introduce logic for parsing the partitions in the partial removal scenario. In that case I would show the warning.
shall we decide on the warning content?

adamkankovsky (approved):
Seems fine to me. Thank you very much for the changes, it improved the readability of the code itself.

PR #340, opened 2024-07-04
tests: move reclaim space tests into seperate file

KKoukiou:


adamkankovsky (approved):
LGTM, It is true that there were already too many tests in the check-storage.

PR #341, opened 2024-07-04
storage: fix available space in reclaim dialog when parent device is deleted

KKoukiou:


src/components/storage/ReclaimSpaceModal.jsx
```diff
@@ -129,9 +131,15 @@ const ReclaimFooter = ({ isFormDisabled, onClose, onReclaim, unappliedActions })
     const { devices } = deviceTrees[""];
     const diskFreeSpace = useDiskFreeSpace({ devices, selectedDisks: diskSelection.selectedDisks });
     const requiredSize = useRequiredSize();
+    const isDeviceRemoved = device => (
+        unappliedActions[device].includes("remove")
+    );
+    const isDeviceParentRemoved = device => (
+        getDeviceAncestors(devices, device).some(parent => unappliedActions[parent].includes("remove"))
```
martinpitt, line 138: This would be a little more elegant/less DRY with `.some(isDeviceRemoved)`

src/helpers/storage.js
```diff
@@ -22,12 +22,11 @@
  * @returns {Array}
  */
 export const getDeviceAncestors = (deviceData, device) => {
-    // device ancestors including the device itself
     const ancestors = [];
     const deviceParents = deviceData[device]?.parents?.v || [];
 
-    ancestors.push(device);
     deviceParents.forEach(parent => {
+        ancestors.push(parent);
```
martinpitt, line 29: The impact of this change isn't obvious from just the patch, it depends on where else `deviceParents` is used. If you checked this, ok.

martinpitt (approved):
Thanks! Looks good enough, happy to re-review if you want to clean this up.

src/helpers/storage.js
KKoukiou: Yeap I checked it.

martinpitt (approved):
Cheers!

PR #342, opened 2024-07-04
tests: storage: convert most of the storage tests to non destructive

KKoukiou:
Most of them were already 'nondestructive', the ones which were not is because they were performing some destructive storage operations. Since we don't share the target disk between the tests any more, we are safe to mark these as non-destructive.


martinpitt (approved):
Nice!

PR #348, opened 2024-07-10
review: show resized partitions that belong to an OS

KKoukiou:
- [x] Blocked on new blivet release containing https://github.com/storaged-project/blivet/pull/1256 in rawhide

### DIFF
--- a/src/components/review/StorageReview.jsx
+++ b/src/components/review/StorageReview.jsx
@@ -214,64 +214,120 @@ const isDeviceDeleted = ({ actions, device }) => (
     actions.find(action => action["device-name"].v === device && action["action-type"].v === "destroy")
 );
 
-export const StorageReviewNote = () => {
-    const plannedActions = usePlannedActions();
+/**
+ * @returns {boolean}   True is the device will be resized according to the actions
+ */
+const isDeviceResized = ({ actions, device }) => (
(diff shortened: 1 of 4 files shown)


src/components/review/StorageReview.jsx
```diff
@@ -214,65 +214,119 @@ const isDeviceDeleted = ({ actions, device }) => (
...
-        system => (
-            system.devices.v.some(device => isDeviceDeleted({ actions: plannedActions, device })) &&
-            !system.devices.v.every(device => isDeviceDeleted({ actions: plannedActions, device }))
-        )
+        system => {
+            const systemDevices = system.devices.v;
+
+            return (
+                systemDevices.some(check) &&
+                !systemDevices.every(check)
```
rvykydal, line 256: If all the devices of a system are resized (not saying it makes sense), would the system be reported as affected?

KKoukiou: Oh it will not. I had this check for 'deleted systems', I report in the review screen systems which were comletely wiped out differently than ones with partitially deleted partitions. 

For resize this is not interesting. I guess I need to adjust the (!systemDevices.every) to do that only for deleted.

KKoukiou: Fixed

rvykydal (approved):
Looks good to me.

PR #365, opened 2024-07-24
tests: map test matrix from Fedora QA team to our tests

KKoukiou:
https://fedoraproject.org/wiki/Test_Results:Fedora_41_Rawhide_20240706.n.1_Installation#Test_Matrix

### DIFF
--- a/test/check-storage-basic
+++ b/test/check-storage-basic
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program; If not, see <http://www.gnu.org/licenses/>.
 
-from anacondalib import VirtInstallMachineCase, pixel_tests_ignore
+from anacondalib import VirtInstallMachineCase, pixel_tests_ignore, test_plan
 from installer import Installer
 from password import Password
 from progress import Progress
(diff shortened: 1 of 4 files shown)


test/check-storage-basic
```diff
@@ -304,53 +315,95 @@ class TestStorage(anacondalib.VirtInstallMachineCase, StorageHelpers):
...
-        s.check_disk_visible("vda")
-        s.check_disk_visible(dev)
+        m.reboot()
+        self.selectBootMenuEntry(2)
+        # FIXME: Debian will not boot unless we remove the vda disk
+        # pretty_name = get_pretty_name(m)
+        # self.assertIn("Debian GNU/Linux", pretty_name)
 
-        s.wait_no_disks_detected_not_present()
+    @test_plan("https://fedoraproject.org/wiki/QA:Testcase_partitioning_guided_multi_empty_all")
```
rvykydal, line 381: I don't see this Expected result tested here:
"The created partition layout should span all the target devices."

rvykydal: I wonder if it actually belongs under this expected result:
"The expected partition layout should be created on the target device(s)."
which is not checked as well, but maybe on purpose.

rvykydal:
So the mapping mechanism looks good to me.

Perhaps the webui tests are not expected to test the "Expected Results" of the QA tests they map to, or not completely, but I think for example "The created partition layout should span all the target devices." seems like a valid check to me (or is it maybe implicit by the fact the OS can be installed bootable only if both disks are actually used?)

KKoukiou:
> So the mapping mechanism looks good to me.
> 
> Perhaps the webui tests are not expected to test the "Expected Results" of the QA tests they map to, or not completely, but I think for example "The created partition layout should span all the target devices." seems like a valid check to me (or is it maybe implicit by the fact the OS can be installed bootable only if both disks are actually used?)

OK, so I extended 3 of the mapped tests, to do the layout verification with lsblk. I removed the rest of the mappings, as the other tests where not e2e, therefore this layout verification was not possible. 
Let's think more about it this / next week. 
Leaning towards indeed covering all the Test Plans from Fedora QA with e2e tests. If these end up taking too long to complete, we can introduce a way to run in PRs 'non e2e' tests and only before merging to trigger the whole suite. (Similar to kickstart tests)

rvykydal (approved):
Looks good to me.

PR #371, opened 2024-07-26
tests: introduce expensive scenario for the Fedora Plan tests

KKoukiou:


martinpitt:
Looks okay, but what's up with [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-371-359bfec7-20240726-142801-fedora-rawhide-boot-expensive-rhinstaller-anaconda-webui/log.html)?

KKoukiou:
> Looks okay, but what's up with [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-371-359bfec7-20240726-142801-fedora-rawhide-boot-expensive-rhinstaller-anaconda-webui/log.html)?

Yes tests are not picked up correctly. Drafting again.

KKoukiou:
@martinpitt weirdly although tests run correctly and there are not failures, (there is 1 retry which should be accepted) the container exited with error code. Do you have any idea why?

martinpitt:
Ah, I figured you spotted the 

> .cockpit-ci/run: line 48: unexpected EOF while looking for matching ``'

in the raw log. That was indeed a bit hidden.

Makefile
```diff
@@ -198,12 +198,6 @@ create-updates.img: bots
 	-rm $(UPDATES_IMG)
 	make $(UPDATES_IMG)
 
-# test runs in kernel_t context and triggers massive amounts of SELinux
-# denials; SELinux gets disabled, but would still trigger unexpected messages
-# we create huge VMs, so we need to reduce parallelism on CI
-integration-test: prepare-test-deps test/reference $(UPDATES_IMG)
```
martinpitt, line 204: This means there's no longer a single `make ...` which you can use in a local checkout to run integration tests. But I figure if you wanted that, inspecting or running `test/run` should actually do fine.

test/run
```diff
@@ -1,8 +1,48 @@
...
+# denials; SELinux gets disabled, but would still trigger unexpected messages
+# we create huge VMs, so we need to reduce parallelism on CI
+if [ -z "${TEST_JOBS-}" ]; then
+  TEST_JOBS=4
+fi
+if [ -z "${TEST_OS-}" ]; then
+  TEST_OS=fedora-rawhide-boot
+fi
+export TEST_JOBS TEST_OS
+J=$(($TEST_JOBS/4)); [ $J -ge 1 ] || J=1; TEST_AUDIT_NO_SELINUX=1 test/common/run-tests --test-dir test/ --jobs $J $RUN_OPTS
```
martinpitt, line 48: If `$TEST_JOBS` isn't defined, such as outside of our CI environment, this will break with a syntax error. But that is already true in main, it's just moved. Just mentioning because it'd break running `test/run` locally after a checkout.

martinpitt (approved):
Thanks! Just a minor thing that can be fixed easily as a follow-up if it bothers you.

PR #372, opened 2024-07-26
test: we dont import whole anacondalib module

KKoukiou:


KKoukiou:
This slipped because 2 PRs where merged that introduced incompatible changes. Test where passing because they did not have merge conficts.

rvykydal (approved):

martinpitt:
Ah, that explains https://github.com/cockpit-project/bots/pull/6664 . This is so obvious and main is broken, there can't possibly be any objection to this. Thanks!

PR #377, opened 2024-08-02
infra: run all tests when triggered with non bots defined scenario

KKoukiou:
The are a few more TEST_SCENARIOS values aside from 'expensive' and 'other'. Such an example are the anaconda-pr-* and cockpit-pr-* scenarios, which are defined when we trigger Web UI tests from anaconda core and cockpit repositories respectively.

### DIFF
--- a/test/run
+++ b/test/run
@@ -28,12 +28,10 @@ case "${TEST_SCENARIO:=}" in
     *other*)
         RUN_OPTS="$(echo "$ALL_TESTS" | grep -Ev "$RE_EXPENSIVE")"
         ;;&
+    *)
+        RUN_OPTS="$ALL_TESTS"
+        ;;&
 esac
-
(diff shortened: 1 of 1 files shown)


jkonecny12:
I wonder, where it will fail if the scenario is not supported when the check code was removed?

KKoukiou:
> I wonder, where it will fail if the scenario is not supported when the check code was removed?

Well it will not. It will just run all tests instead of failing by default. 
Test scenarios are hardcoded in the repos, (anaconda, cockpit, bots), so they are not changing unpredictably.

jkonecny12:
I wonder how problematic that could be. In general, if you do a typo it will run something you don't really want which I see as bad UX.

Anyway, it's for testing purpose and could be adjusted later. Let's not block on this if you don't want to.

jkonecny12 (approved):

PR #380, opened 2024-08-05
storage: increase the radio button spacing in the scenario selection group

KKoukiou:
After designer feedback.

### DIFF
--- a/src/components/storage/InstallationScenario.jsx
+++ b/src/components/storage/InstallationScenario.jsx
@@ -429,7 +429,7 @@ export const InstallationScenario = ({
     return (
         <>
             <Title headingLevel={headingLevel}>{_("How would you like to install?")}</Title>
-            <FormGroup isStack hasNoPaddingTop data-scenario={storageScenarioId}>
+            <FormGroup className={idPrefix + "-scenario-group"} isStack hasNoPaddingTop data-scenario={storageScenarioId}>
                 <InstallationScenarioSelector
                   dispatch={dispatch}
                   idPrefix={idPrefix}
--- a/src/components/storage/InstallationScenario.scss
+++ b/src/components/storage/InstallationScenario.scss
@@ -13,3 +13,8 @@
     }
   }
 }
+
+// Increase the standard radio button spacing
+.installation-method-scenario-group .pf-v5-c-form__group-control {
+  gap: var(--pf-v5-global--spacer--md);
+}


garrett:
## Summary

Thanks! I think it's already better, but that we _probably_ need an `lg` spacer, not an `md` one.

## Detail

There's no screenshot included, so I dug around and looked for a screenshot from the tests. If this is the correct screenshot, it's _probably_ better even if it's not as much spacing as in the mockups.

I think to be consistent with the mockups, it'd need a spacer of `lg` instead of `md`, as there seems to be around 6px to 8px difference (which is hard to ascertain, due to different radio button graphics due to being in Chrome vs. Firefox vs. the mockups, so I don't have exact px; it could be around 2px off). The difference going from `md` to `lg` is around that same amount (`8px`, as it's from `md` @ `16px` to `lg` @ `24px`).

It's also hard to judge visually with just two options.

This is the screenshot I saw in the tests:

image
from https://github.com/rhinstaller/pixel-test-reference/compare/4d19796b9fc4f429e2866901fbec2f9ee7214b12...344c6dc7b22288ece70ba707e6b996302714f565#diff-03fe8a0953de24d77499097d554892b6101dde40520127c9c3c009af5bb91e3c

And here's a difference overlay @ 50% opacity, so the screenshot is 100% and the mockup is 50% difference:

image

Lots of the spacing is different (which is somewhat expected); it's aligned to the baseline of the text of the first option. However, as mentioned above, the radio buttons as well as other factors may mean this could be a pixel or two off. But it's definitely smaller in space than the mockup, and lg might be better than md @ https://github.com/rhinstaller/anaconda-webui/pull/380/files#diff-9581ddeb3874fc78679866ad90e57039d2dd413b83e9903768a6dd211093a03bR19.

KKoukiou:
@garrett the screenshots (pixel test updates) can be seen in the commit changes. https://github.com/rhinstaller/pixel-test-reference/compare/4d19796b9fc4f429e2866901fbec2f9ee7214b12..344c6dc7b22288ece70ba707e6b996302714f565 here is the delta. 

So should I update with LG spacing?

garrett:
> So should I update with LG spacing?

Yes, please. Thanks!

If we think it's too much in practice, we can dial it back. But I think it might be right.

garrett (approved):
Looks good.

We'll catch it later in context if it needs any more adjustments (with more installation methods, the reclaim option, changed text), but let's go with this.

PR #381, opened 2024-08-06
tests: fix test scenario picking always all tests

KKoukiou:
Dont move to next condition if we matched one already. Commit e04d939dbfabd5aec357ca108395d40ae72fd7fa incorrectly introduced a new case, which matched always because of the ;;$.

From man bash:

Using ;;& in place of ;; causes the shell to test
the next pattern list in the statement, if any, and execute any associated list on a successful match.


marusak (approved):
So these are exclusive, right?

KKoukiou:
> So these are exclusive, right?

These are exclusive yes.

PR #384, opened 2024-08-06
Reduce miliseconds for debouncing validity checks after user input

KKoukiou:
Resolves: rhbz#2280016

### DIFF
--- a/src/components/Password.jsx
+++ b/src/components/Password.jsx
@@ -137,11 +137,11 @@ export const PasswordFormFields = ({
     const [passwordStrength, setPasswordStrength] = useState("");
 
     useEffect(() => {
-        debounce(300, () => setCheckPassword(password))();
+        debounce(100, () => setCheckPassword(password))();
     }, [password, setCheckPassword]);
 
     useEffect(() => {
-        debounce(300, () => setCheckConfirmPassword(confirmPassword))();
+        debounce(100, () => setCheckConfirmPassword(confirmPassword))();
     }, [confirmPassword, setCheckConfirmPassword]);
 
     const ruleResults = useMemo(() => {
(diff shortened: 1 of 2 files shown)


garrett:
Having a shorter debounce time for a quick typist can help, for sure, but...

I'm not sure if the complaint is that the check isn't run quick enough, however. Perhaps what they're really asking for is for some friction to prevent using simpler passwords?

The thing is, the weak password check is already done on the first password entry, and there is a "weak" indicator in the top-right. It might not be as obvious of a placement as underneath, where there's already checker feedback.

luks_passphrase-2024-05-10_17 01 00∶00∶12

Here's a mockup that not only uses proper spacing between passphrase and confirm passphrase (now mentioned in our document), but it shows what it would look like if we made the warning actually look like a warning (instead of an error) and had it in the check area instead (so it's within the line of sight of the checks, and comes _after_ the item that's being typed instead of _before_):

Wizard, encryption, selected(3)


Please note that this is _different_ from what PatternFly suggests (as they have this as an error in the top right instead, which is what Anaconda is doing, even though we're not treating it as an error and allowing it regardless). I think, in this case, PatternFly is doing it wrong, and the strength message should be shown in the area with the other qualifiers. Especially so since we have a confirmation entry (which PatternFly doesn't suggest), and our confirmation entry is directly below. So the order is do something, see the results (if it passes, fails, or has a warning), do something (repeat password), see the result (if it matches or not). Instead of what it's like currently, where you have to look before what you do to see if it's fine or not.

garrett:
Moving the warning to a more obvious place like in the mockup would help as well as changing the bounce timing (which should be instant when blurring and only be a debounce when doing a keyup).

But it doesn't add more friction, like having a modal or something to confirm it... which could be annoying, especially since there's already friction to re-type the password, which is supposed to help prevent the weak passwords by giving someone time to see that. We don't let someone proceed without the retyping, so I think we don't need anything more than to make sure it's obvious while they're retyping (or copy/pasting the password if they have it unhidden and can see it's what they expect).

In other words: I think functionally, we don't need to change the interaction by adding a modal or some opt-in to go forward with an insecure password. We have that already with "repeat".

PR #388, opened 2024-08-07
storage: reclaim: don't schedule actions for partitions whose parents are removed

KKoukiou:


adamkankovsky (approved):
I have tried it and it seems to work as it should.

PR #391, opened 2024-08-08
storage: some enhancements on reclaim dialog after designer feedback

KKoukiou:
New
127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html (3)

Old
regular-reclaim-dialog-table


garrett (approved):
Definitely better!

It still has some issues (too much space for the disks vs. the rest, sizes and actions aren't right aligned, etc.), but this is definitely a step in the right direction.

PR #392, opened 2024-08-09
storage: reclaim dialog: show format type not device type

KKoukiou:
This needs to be improved with human readable strings for partition labels once https://github.com/storaged-project/blivet/issues/1258 is implemented.

Windows (missing FS types because it's mocked disk)
TestReclaimExistingSystemWindows-testBasic-reclaim-space-modal-windows-pixels

Fedora
TestReclaimExistingSystemFedora-testDeletePartition-reclaim-space-modal-fedora-pixels

### DIFF
--- a/src/actions/storage-actions.js
+++ b/src/actions/storage-actions.js
@@ -69,6 +69,15 @@ export const getDevicesAction = () => {
                     const formatData = await getFormatData({ diskName: device });
                     devData.formatData = formatData;
 
+                    // FIXME: https://github.com/storaged-project/blivet/issues/1258
+                    // This is a workaround for the issue with missing partition label human readable strings in blivet
+                    const partName = await cockpit.script(`udevadm info --query=property --name=${devData.path.v} | grep PARTNAME= | cut -d= -f2`);
+                    const fsLabel = await cockpit.script(`udevadm info --query=property --name=${devData.path.v} | grep ID_FS_LABEL= | cut -d= -f2`);
+                    devData.misc = {
+                        fsLabel: fsLabel.trim(),
+                        partName: partName.trim(),
+                    };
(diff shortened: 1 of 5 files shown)


garrett:
This is what I've had to show different OSes:

partitions, windows + linux + data, 2 disks(4)

Location is the device/partition. Name is the human readible name (EFI, Windows partition, recovery, Boot, Fedora partition with its subvolumes, Swap, Free space), type (filesystem), space, and the actions.

The disk name is so wide, it spans both location and name with a colspan.

Subvolumes use the same location as the volume; they're dropped out in the UI as that's just otherwise noise, and they're indented under the name to make it more obvious that they're children of the volume. (root and home being btrfs subvolumes under the Fedora installation)

garrett:
Here's a visualization of the table, to show what I mean about the colspan for the disk name:

image

(I was using a grid which has fr sizing, so ignore the spacing to the right of the device/partition location column; it added extra space inside the area, which wouldn't be done like that with a table.)

garrett:
Basically, it's OS for the volume and types for subvolumes. But if it's using only partitions that aren't subvolumes, then it would be a string of "OS name" + " " + "partition type". For example, same thing, but with CentOS having / (root) and /home:

partitions, windows + linux + data, 2 disks(5)

And that's what the "Windows" and "Windows recovery" are about as well.

KKoukiou:
@garrett 
``` But if it's using only partitions that aren't subvolumes, then it would be a string of "OS name" + " " + "partition type". ```
That's too much fine grained testing to do in the UI for my taste.

I would go for the first pass at least without this logic.

adamkankovsky (approved):
LGTM, thanks.

PR #393, opened 2024-08-09
tests: test reclaim from encrypted LUKS partitions spanning two disks on LVM VG

KKoukiou:
Blocked by https://issues.redhat.com/browse/INSTALLER-4011


KKoukiou:
We agreed that for encrypted vgs we leave it up to the user.

PR #395, opened 2024-08-09
tests: using one disk from LVM VG spanning two disks is prohibited

KKoukiou:


KKoukiou (commented):
The error is correct but this should appear earlier.

M4rtinK (approved):
Looks good to me, thanks! :)

PR #398, opened 2024-08-12
Allow having duplicate device names

KKoukiou:
Since we are addressing devices through their device-id it's now possible to utilize for installation devices with the same name.


adamkankovsky (approved):
I've tested it, probably some more tests could be added for more complex interactions with this kind of storage, but it looks good to me

KKoukiou:
> I've tested it, probably some more tests could be added for more complex interactions with this kind of storage, but it looks good to me

I might add some in the future. Though it's more of an edge case scenario - I dont think we need to waste CI resources on more tests.

PR #399, opened 2024-08-12
reclaim: Make the list scroll instead of the modal body

garrett:
When there isn't enough vertical height to show all the partitions (when there are a lot of partitions), make the list of partitionsscroll instead of the body of the modal.

Screenshot, where I attached a Windows VM as a disk image and set the height to be artificially short as to trigger the overflow:

image

Note the scrollbar on the right, as it is attached to the list instead of the modal.

### DIFF
--- a/src/components/storage/ReclaimSpaceModal.scss
+++ b/src/components/storage/ReclaimSpaceModal.scss
@@ -21,7 +21,7 @@
   }
 
   tr.reclaim-space-modal-table-row.reclaim-space-modal-device-level-0 {
-    border-block-start:  var(--pf-v5-c-table--border-width--base) solid var(--pf-v5-c-table--BorderColor);
+    border-block-start: var(--pf-v5-c-table--border-width--base) solid var(--pf-v5-c-table--BorderColor);
   }
 
   .reclaim-space-modal-device-action-delete {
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
For the sticky header, Patternfly suggests wrapping with:
OuterScrollContainer and InnerScrollContainer and applying isStickyHeader property.

If we can drop this custom CSS in favor of the patternfly property that would be ideal.

Take in mind that when you develop, you can update the VM with `RSYNC=test-updates make watch`.
You need to define the host like this: https://github.com/rhinstaller/anaconda-webui/tree/main/test#helpful-tips

Feel free to use the name of your preference.

garrett:
>  For the sticky header, Patternfly suggests wrapping with:
OuterScrollContainer and InnerScrollContainer and applying isStickyHeader property.

Sure! If you'd rather use that and it provides the same effect, go for it!

KKoukiou:
> > For the sticky header, Patternfly suggests wrapping with:
> > OuterScrollContainer and InnerScrollContainer and applying isStickyHeader property.
> 
> Sure! If you'd rather use that and it provides the same effect, go for it!

I would prefer if you try that out. I tried to add that and the overflows where messed up.

KKoukiou (approved):
Merging this as it's a nice fix, and I dont have capacity to deal with the suggested enhancements.
@garrett maybe when you re back, can you take a look and send a follow up?

KKoukiou (approved):

garrett:
>   @garrett maybe when you re back, can you take a look and send a follow up?

This PR is feature complete and working as intended. I don't know how to make React do this, and Anaconda VMs are currently not working again. (I don't know why. Other VMs work; this is specifically an Anaconda WebUI issue.)

I don't think there is an "easy" way to make PF-React do the right thing here, as it needs to adjust the box body, stack, panel, and table head and do so by changing the layout with respect to flex and overflows.

In addition, setting sticky as a React property won't work for the heading, unless that also adjusts the color (as it needs an outline and background too). And speaking of style, the panel needs a border as well after adjusting sizing.

If you happen to know of a way to make it work, and if PF-React actually does all this stuff with some specific properties, then that'd be awesome. But I highly doubt that it can.

garrett:
>   @garrett maybe when you re back, can you take a look and send a follow up?

This PR is feature complete and working as intended. I don't know how to make PF-React do this.

I don't think there is an "easy" way to make PF-React do the right thing here, as it needs to adjust the box body, stack, panel, and table head and do so by changing the layout with respect to flex and overflows.

In addition, setting sticky as a React property won't work for the heading, unless that also adjusts the color (as it needs an outline and background too) — but perhaps it does do this. And speaking of style, the panel needs a border as well after adjusting sizing.

If you happen to know of a way to make it work, and if PF-React actually does all this stuff with some specific properties, then that'd be awesome. But I highly doubt that it can.

Lastly, I spent all day so far to set up a VM (which is now working), and then trying to re-implement parts of this in PF-React, but I can't figure it out. The table properties you talk about are specific to a PatternFly table and add wrappers (for some unknown reason as I was able to do that without any of that in the CSS), but we're using a Cockpit table instead, and I can't get it working.

I could've spent the two half days I struggled with trying to get this in React in something else design related, which is needed, instead of attempting to try to convince React to "do the right thing" here.

KKoukiou (approved):

PR #401, opened 2024-08-14
test: Move from CDP to BiDi

KKoukiou:
See cockpit-project/cockpit#20832

Drop chrome-remote-interface NPM dependency.

Explicitly add a "glob" devDependency to follow suit with cockpit-project/cockpit@680decc155a
It was previously used implicitly through a transient dependency of something else, but our esbuild po-plugin uses it explicitly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,12 +114,9 @@ install: $(DIST_TEST) po/LINGUAS
 	cp webui-desktop $(DESTDIR)/usr/libexec/
 	ln -sTfr $(DESTDIR)/usr/share/pixmaps/fedora-logo-sprite.svg $(DESTDIR)/usr/share/cockpit/$(PACKAGE_NAME)/logo.svg
 
-# required for running integration tests; commander and ws are deps of chrome-remote-interface
+# required for running integration tests;
 TEST_NPMS = \
-	node_modules/chrome-remote-interface \
-	node_modules/commander \
 	node_modules/sizzle \
-	node_modules/ws \
 	$(NULL)
 
 dist: $(TARFILE)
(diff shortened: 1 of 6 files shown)


KKoukiou (commented):
@martinpitt i have two tests that check exception handling fail with this bidi convertion.

both do something like: 253             b.eval_js("window.setTimeout(function() { myNonExistingFunction()}, 0);")              


And because the UI handles unhandled JS errors with window.onerror I expect to see the error dialog.

With bidi this does not propagate properly. Any ideas what's missing? 

Otherwise I will disable the two tests.

martinpitt:
Some notes:
 - The old CDP API didn't have reliable exception reporting, it was async and delayed. That was the reason why you had to do these loops:
```py
  b.eval_js("window.setTimeout(function() { myNonExistingFunction()}, 0);"
  # Some round trips, one of which should update the deferred exception
  for _ in range(0, 5):
      b.wait_in_text("#critical-error-bz-report-modal-details", "myNonExistingFunction is not defined")
      time.sleep(2)
```
With BiDi, this will happen immediately and reliably *if* the command or returned promise rejects. However, the called command does *not* fail, as it just starts a timer and does not return a promise. So you can drop the loop (as it's pointless), and drop the `with self.assertRaises(RuntimeError):` as the eval_js() itself will reliably *not* fail. You only get the console error:

> error: ReferenceError: myNonExistingFunction is not defined

from the timer when it fires, but as nothing pays attention to that, this doesn't cause any crash.

- Calling the timeout crash fn did *not* trigger Cockpit's general "Oops" handler with CDP either. This happens in CDP context, not page context. Supposedly Anaconda's error handler uses a different approach than cockpit's Oops handler.

We test that in cockpit in https://github.com/cockpit-project/cockpit/blob/main/pkg/playground/exception.js , and the test clicks that button to provoke a crash. Taking that idea, I was trying to do this not with a timer, but with setting an event handler inside of the actual page:

```
b.eval_js("window.document.getElementById('toolbar').addEventListener('click', () => myNonExistingFunction())")
b.click("#toolbar")
```
Doing that in the JS console/inspector and clicking the toolbar works fine, but not through `eval_js`. Apparently this is running in a "shielded" context. 

@KKoukiou Sorry, I'm afraid this is a hard problem. My suggestion would be to add that crasher into the actual page somehwere, in a hidden place (making it conditional on running as test or so?), and then activate it with a click.

KKoukiou:
@martinpitt I just realized that the test correctly triggers the exception.  I see the exception handling UI in the screenshots. The problem is that RuntimeError was never raised, aka - removing ` with self.assertRaises(RuntimeError):` fixes the issue. Thanks for taking a look.

marusak (approved):
Thanks!

M4rtinK (approved):
Looks good to me. :)

martinpitt:
@KKoukiou this looks great now, thanks! I'm glad it was something easy after all.

PR #408, opened 2024-08-19
storage: mount point assignment: show partition labels in device selection

KKoukiou:
Updated UI:
127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html

### DIFF
--- a/src/components/storage/MountPointMapping.jsx
+++ b/src/components/storage/MountPointMapping.jsx
@@ -264,10 +264,20 @@ const DeviceColumnSelect = ({ deviceData, devices, handleRequestChange, idPrefix
     const size = cockpit.format_bytes(deviceData[device]?.total.v);
     const options = devices.map(device => {
         const deviceName = deviceData[device].name.v;
+
+        const ancestors = getDeviceAncestors(deviceData, device);
+        const parentPartition = [device, ...ancestors].find(ancestor => deviceData[ancestor].type.v === "partition");
+        const typeLabel = deviceData[parentPartition]?.misc.fsLabel || deviceData[parentPartition]?.misc.partName;
+
         const formatType = deviceData[device]?.formatData.type.v;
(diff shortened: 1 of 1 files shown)


src/components/storage/MountPointMapping.jsx
```diff
@@ -264,10 +264,20 @@ const DeviceColumnSelect = ({ deviceData, devices, handleRequestChange, idPrefix
     const size = cockpit.format_bytes(deviceData[device]?.total.v);
     const options = devices.map(device => {
         const deviceName = deviceData[device].name.v;
+
+        const ancestors = getDeviceAncestors(deviceData, device);
+        const parentPartition = [device, ...ancestors].find(ancestor => deviceData[ancestor].type.v === "partition");
```
adamkankovsky, line 269: Can't there be a situation where the parent partition is null? Because that could cause an error

KKoukiou: Do you mean that: deviceData[ancestor] is undefined?

Well ancestors should be always an array - sometimes an empty one.

adamkankovsky: Because in the following code we feel that we always get the parentPartition, so if there can't be some edge case in which this doesn't work

KKoukiou: Hm I see what you mean now.

I really cant imagine how we can have a storage layout where there is not a partition in the device tree.

however you 're right - we shall be better be safe - will adjust.

KKoukiou: Did the safe access change.

rvykydal (approved):
LGTM, very nice.

rvykydal:
Sharing screenshot for our ubuntu-stable testing image
Screenshot from 2024-09-18 09-33-45

PR #409, opened 2024-08-19
Split storage tests into seperate files according to storage technology

KKoukiou:
Do this for better organization and repository readability.


adamkankovsky (approved):
This is a very good idea, thanks for it.

PR #417, opened 2024-08-26
tests: ignore changes in the payload size for pixel tests

KKoukiou:


src/components/storage/ReclaimSpaceModal.jsx
```diff
@@ -203,11 +203,11 @@ const ReclaimFooter = ({ isFormDisabled, onClose, onReclaim, unappliedActions })
     return (
         <Stack hasGutter>
             <HelperText>
-                <HelperTextItem isDynamic variant={status}>
+                <HelperTextItem id={idPrefix + "-hint"} isDynamic variant={status}>
                     {fmtToFragments(
                         _("Available free space: $0. Installation requires: $1."),
                         <b id={idPrefix + "-hint-available-free-space"}>{cockpit.format_bytes(diskFreeSpace + selectedSpaceToReclaim)}</b>,
-                        <b>{cockpit.format_bytes(requiredSize)}</b>
+                        <b id={idPrefix + "-hint-required-free-space"}>{cockpit.format_bytes(requiredSize)}</b>
```
martinpitt, line 210: So this ID is unused. You now ignore the entire hint, not just the size. I guess because the text after `$0` will shift according to the width of requiredSize?

martinpitt (approved):
Could be reduced, but fine, thanks!

src/components/storage/ReclaimSpaceModal.jsx
KKoukiou: Yes right, and for RTL it shifts even the previous string, so I had to ignore the whole line

PR #421, opened 2024-08-28
Change tarball payload for rpm

adamkankovsky:


test/prepare-updates-img
```diff
@@ -14,6 +14,9 @@ for i in ${TEST_RPMS//,/ }; do
     cp $i tmp/rpms
 done
 
+# Copy the payload RPM to the tmp/rpms
+cp "$(ls bots/images/fedora-rawhide-anaconda-payload.rpm)" tmp/rpms/
```
KKoukiou, line 18: This will not read the symlink path: you need: readlink -f bots/images/fedora-rawhide-anaconda-payload

KKoukiou: @adamkankovsky can you please fix this according to the suggestion? Without readlink we dont get the symlink path

test/machine_install.py
```diff
@@ -83,11 +83,6 @@ def _serve_payload(self):
 
         return payload_cached_name, http_payload_port
 
-    def _write_interactive_defaults_ks(self):
-        payload_cached_name, http_payload_port = self._serve_payload()
```
KKoukiou, line 87: You can also remove the serve_payload function now

test/prepare-updates-img
KKoukiou: Looks like cp -L bots/images/fedora-rawhide-anaconda-payload.rpm tmp/rpms/
works

adamkankovsky: cp -L unfortunately doesn't work as it should

test/prepare-updates-img
```diff
@@ -14,6 +14,9 @@ for i in ${TEST_RPMS//,/ }; do
     cp $i tmp/rpms
 done
 
+# Copy the payload RPM to the tmp/rpms
+cp "$(readlink -f bots/images/fedora-rawhide-anaconda-payload)" tmp/rpms/
```
KKoukiou, line 18: You can also do: `cp -L bots/images/fedora-rawhide-anaconda-payload tmp/rpms/fedora-rawhide-anaconda-payload.rpm`

Feel free top decide, I think the new suggestion is cleaner.

KKoukiou:
Looks like this was not the correct approach https://github.com/cockpit-project/bots/pull/6801

PR #424, opened 2024-08-30
[WIP] Home reuse poc

rvykydal:
Requires https://github.com/rhinstaller/anaconda/pull/5814

TODO:
- [ ] skip encryption page (instead of disabling checkbox which is a workaround)
- [ ] add support for unlocking (checkbox similar to reclaim dialog?)
- [ ] polish review screen
- [ ] add destructive e2e test testing booting and installed system
- [ ] run and fix tests


src/apis/storage_partitioning.js
```diff
@@ -95,6 +95,19 @@ export const partitioningSetEncrypt = ({ encrypt, partitioning }) => {
             });
 };
 
+/**
+ * @param {string} partitioning     DBus path to a partitioning
+ */
+export const partitioningSetHomeReuse = ({ partitioning }) => {
+    return getPartitioningRequest({ partitioning })
```
KKoukiou, line 102: I think this is nicer writen with async await. however I know we still use a lot of then so not objection if you want to keep it like this.

For the sake of it, that would be 

``` js
export const partitioningSetHomeReuse = async ({ partitioning }) => {
  const request = await getPartitioningRequest({ partitioning });
  ....
  await setPartitioningRequest({ partitioning, request });;
};
```

src/components/review/StorageReview.jsx
```diff
@@ -86,11 +87,15 @@ const DeviceRow = ({ disk }) => {
         const size = cockpit.format_bytes(devices[device].size.v);
         const request = requests.find(request => request["device-spec"] === device);
         const format = devices[device].formatData.type.v;
-        const action = (
+        let action = (
             request === undefined || request.reformat
                 ? (format ? cockpit.format(_("format as $0"), format) : null)
                 : ((format === "biosboot") ? format : _("mount"))
         );
+        // FIXME use better JS
```
KKoukiou, line 95: It's indeed weird to do the inline conditional and then do the if conditional to possibly overwrite it.
Since we have many conditions here, let's do them all in if clauses.

src/components/storage/Common.jsx
```diff
@@ -143,7 +144,11 @@ export const getNewPartitioning = async ({
 
     const part = await createPartitioning({ method });
 
-    if (currentPartitioning?.method === method && method === "AUTOMATIC" && currentPartitioning.requests[0].encrypted) {
+    if (storageScenarioId === "home-reuse") {
+        await partitioningSetHomeReuse({ partitioning: part });
+    }
+
+    if (currentPartitioning?.method === method && method === "AUTOMATIC" && storageScenarioId !== "home-reuse" && currentPartitioning.requests[0].encrypted) {
```
KKoukiou, line 151: Can you do a multiliner with this please? 
It's too long now.

src/components/storage/DiskEncryption.jsx
```diff
@@ -194,6 +194,7 @@ export class Page {
     constructor (isBootIso, storageScenarioId) {
         this.component = DiskEncryption;
         this.id = "disk-encryption";
+        // TODO hide for home-reuse
         this.isHidden = ["mount-point-mapping", "use-configured-storage"].includes(storageScenarioId);
```
KKoukiou, line 199: Just add the scenario id here in the list.

test/check-storage-home-reuse
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2022 Red Hat, Inc.
```
KKoukiou, line 3: s/2022/2024

test/check-storage-home-reuse
```diff
@@ -0,0 +1,96 @@
...
+from storage import Storage
+from storagelib import StorageCase  # pylint: disable=import-error
+from testlib import nondestructive, skipImage, test_main  # pylint: disable=import-error
+
+
+@nondestructive
+class TestHomeReuse(VirtInstallMachineCase, StorageCase):
+    efi = False
+
+    @skipImage("btrfs support missing on fedora-eln image", "fedora-eln-boot")
```
KKoukiou: You can remove this - we dont test eln anymore.

test/check-storage-home-reuse
```diff
@@ -0,0 +1,96 @@
...
+    def testBasic(self):
+        b = self.browser
+        m = self.machine
+        i = Installer(b, m)
+        s = Storage(b, m)
+        r = Review(b, m)
+
+        disk = "/dev/vda"
+        dev = "vda"
+        tmp_mount = "/tmp/home-reuse-mount-test"
```
KKoukiou: It would be cleaner to use an existing Fedora disk, aka `disk_image = "fedora-rawhide"`

KKoukiou: oh, you want the btrfs layout I guess.

src/components/storage/InstallationScenario.jsx
```diff
@@ -125,13 +129,42 @@ const checkMountPointMapping = ({ mountPointConstraints, selectedDisks, usablePa
     return availability;
 };
 
-// TODO implement
-const checkHomeReuse = ({ selectedDisks }) => {
+const checkHomeReuse = ({ devices, originalExistingSystems, selectedDisks }) => {
```
KKoukiou, line 132: No need to pass as property the devices and originalExistingSystems. 

We have custom hooks for that - to make the code cleaner.

const originalExistingSystems = useOriginalExistingSystems();
const devices = useOriginalDevices();

src/components/storage/InstallationScenario.jsx
```diff
@@ -125,13 +129,42 @@ const checkMountPointMapping = ({ mountPointConstraints, selectedDisks, usablePa
...
     const availability = new AvailabilityState();
 
     availability.hidden = false;
     availability.available = !!selectedDisks.length;
 
+    const isOSOnDisks = (osData, disks) => {
```
KKoukiou, line 138: Can you put this in src/helpers/storage.js? I implemented exactly the same also in another PR, https://github.com/rhinstaller/anaconda-webui/pull/440
So feel free to reuse it?

So something like this in src/helpers/storage
``` js
const getOSonDisk = ({ device }) => { 
    const originalExistingSystems = useOriginalExistingSystems();
    const devices = useOriginalDevices();
    const children = getDeviceChildren({ device, deviceData: devices });
    const existingSystemsOnDevice = originalExistingSystems.filter(
        system => system.devices.v.some(device => children.includes(device))
    );
   return existingSystemsonDevice;
}

```

And then here:
```js
    const isOSOnDisks = (disks) => {
    return disks.some(disk => getOSonDisk({ device: disk }).length > 0);
  }
```

src/components/storage/DiskEncryption.jsx
rvykydal: This is not so easy, the disk encryption screen contains application and reset of AUTOMATIC partitioning (which is also /home reuse) so it needs more work (I have it in the TODO list) and maybe some smaller refactoring.

src/components/storage/InstallationScenario.jsx
rvykydal: I was thinking we don't use any hooks / side effects in the checks so I wanted to stick to it also in this check. But it seems these two hooks (using _Original_) don't have any side effects so in this case it does not matter.
Also devices is already passed to checkConfiguredStorage so I wanted to be consistent.

src/components/storage/InstallationScenario.jsx
rvykydal: My function checks if given OS is residing completely on the disks, while the suggested implementation just checks if on the given disks there is any part of any existing OS.

src/components/storage/InstallationScenario.jsx
rvykydal: >  I will use the `useOriginalExistingSystems` in the check

Hm, rather not:
```
#   134:37  error  React Hook "useOriginalExistingSystems" is called in function "checkHomeReuse" that is neither a React function component nor a custom React Hook function. React component names must start with an uppercase letter. React Hook names must start with the word "use"  react-hooks/rules-of-hooks
```

src/apis/storage_partitioning.js
rvykydal: Updated

src/components/review/StorageReview.jsx
rvykydal: Updated

src/components/storage/Common.jsx
rvykydal: Updated

test/check-storage-home-reuse
rvykydal: Updated

src/components/storage/InstallationScenario.jsx
rvykydal: Updated the name of the function

rvykydal:
Added check for partitioning scheme and /home match (from todo list).

rvykydal:
- Updated tests for new disk selection
- Adapted to "erase" -> "reformat" name change.

rvykydal:
@KKoukiou I am trying to address "skip encryption page (instead of disabling checkbox which is a workaround)" from the TODO list in the first comment.
It is the last commit in https://github.com/rhinstaller/anaconda-webui/pull/448 but I am hitting a problem:
The `Next` button of Accounts screen is not updated correctly (it is disabled) when there is already valid information prefilled. See the video where it works for "Erase data and install" where the accounts screen follows the encryption screen but it does not work for "Reinstall with /home partition preserved" where after the "Installation Method screen" the `Next` button on the Accounts screen is disabled:
[next_disabled_on_accounts.webm](https://github.com/user-attachments/assets/235435ee-3063-41cd-8637-40197219f98d)

KKoukiou:
> @KKoukiou I am trying to address "skip encryption page (instead of disabling checkbox which is a workaround)" from the TODO list in the first comment. It is the last commit in #448 but I am hitting a problem: The `Next` button of Accounts screen is not updated correctly (it is disabled) when there is already valid information prefilled. See the video where it works for "Erase data and install" where the accounts screen follows the encryption screen but it does not work for "Reinstall with /home partition preserved" where after the "Installation Method screen" the `Next` button on the Accounts screen is disabled: [next_disabled_on_accounts.webm](https://github.com/user-attachments/assets/235435ee-3063-41cd-8637-40197219f98d)

Will check that in 1h.

src/components/storage/DiskEncryption.jsx
rvykydal: Trying to address in https://github.com/rhinstaller/anaconda-webui/pull/448 with the issues described below https://github.com/rhinstaller/anaconda-webui/pull/424#issuecomment-2367439972

KKoukiou:
> > @KKoukiou I am trying to address "skip encryption page (instead of disabling checkbox which is a workaround)" from the TODO list in the first comment. It is the last commit in #448 but I am hitting a problem: The `Next` button of Accounts screen is not updated correctly (it is disabled) when there is already valid information prefilled. See the video where it works for "Erase data and install" where the accounts screen follows the encryption screen but it does not work for "Reinstall with /home partition preserved" where after the "Installation Method screen" the `Next` button on the Accounts screen is disabled: [next_disabled_on_accounts.webm](https://github.com/user-attachments/assets/235435ee-3063-41cd-8637-40197219f98d)
> 
> Will check that in 1h.

Found that the problem was that goToNext was applied twice:
Fixed by:
``` diff
diff --git a/src/components/storage/InstallationMethod.jsx b/src/components/storage/InstallationMethod.jsx
index 49cd710a..8a3661bb 100644
--- a/src/components/storage/InstallationMethod.jsx
+++ b/src/components/storage/InstallationMethod.jsx
@@ -108,6 +108,7 @@ const CustomFooter = ({ isFormDisabled, isReclaimSpaceCheckboxChecked, setStepNo
     }, [isNextClicked, goToNextStep, newPartitioning, partitioning.path]);
 
     const onNext = async ({ setIsFormDisabled }) => {
         if (method === "MANUAL") {
             setNewPartitioning(partitioning.path);
             setIsNextClicked(true);
@@ -120,11 +121,10 @@ const CustomFooter = ({ isFormDisabled, isReclaimSpaceCheckboxChecked, setStepNo
 
             if (willShowReclaimSpaceModal) {
                 setIsReclaimSpaceModalOpen(true);
-            } else {
+            } else if (storageScenarioId !== "home-reuse") {
                 setIsNextClicked(true);
-            }
-
-            if (storageScenarioId === "home-reuse") {
+            } else {
+                setIsFormDisabled(true);
                 const step = new Page().id;
                 await applyStorage({
                     onFail: ex => {
```

rvykydal:
> Found that the problem was that goToNext was applied twice: Fixed by:
> 

@KKoukiou with the patch the partitioning is reset immediataly after it is applied here: https://github.com/rhinstaller/anaconda-webui/blob/e4a2b9f5dfd6ec982d1fdedced5dbf18a14179b8/src/components/storage/InstallationMethod.jsx#L172
It should be apparent also in the falling [tests](https://cockpit-logs.us-east-1.linodeobjects.com/pull-424-1c09eb93-20240924-090147-fedora-rawhide-boot-anaconda-pr-5814-rhinstaller-anaconda-webui/log.html)
Applying and resetting on the same page seems tricky.

rvykydal:
> I wonder if the same guard as for `use-configured-storage` would work in `needsReset` 

It is in the recent patch, the tests are passing although I am not sure it is correct or acceptable fix.

src/apis/storage_partitioning.js
```diff
@@ -95,6 +95,28 @@ export const partitioningSetEncrypt = ({ encrypt, partitioning }) => {
...
 
+/**
+ * @param {string} partitioning     DBus path to a partitioning
+ */
+export const partitioningSetHomeReuse = async ({ partitioning }) => {
+    const request = await getPartitioningRequest({ partitioning });
+
+    // In general default scheme defined in conf.Storage.default-scheme would be used
+    // Only btrfs (AUTOPART_TYPE_BTRFS) autopartitioning scheme is supported now
+    request["partitioning-scheme"] = cockpit.variant("i", 1);
```
KKoukiou, line 106: Shouldn't the partitioning scheme be set automatically by the backend, similarly how the scheme is currently set by the autopartitioning?

src/apis/storage_partitioning.js
```diff
@@ -95,6 +95,28 @@ export const partitioningSetEncrypt = ({ encrypt, partitioning }) => {
...
+    // These settings work also for AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP schemes
+    request["reformatted-mount-points"] = cockpit.variant("as", ["/"]);
+    request["reused-mount-points"] = cockpit.variant("as", ["/home"]);
+    request["removed-mount-points"] = cockpit.variant("as", ["/boot", "bootloader"]);
+
+    // Settings for AUTOPART_TYPE_PLAIN
```
KKoukiou, line 112: Please dont leave commented out code when this is reaching a final state.

src/components/review/StorageReview.jsx
```diff
@@ -86,11 +87,18 @@ const DeviceRow = ({ disk }) => {
...
         const format = devices[device].formatData.type.v;
-        const action = (
-            request === undefined || request.reformat
-                ? (format ? cockpit.format(_("format as $0"), format) : null)
-                : ((format === "biosboot") ? format : _("mount"))
-        );
+
+        let action = null;
+        if (reuseHomeRequest?.["reused-mount-points"].includes(mount)) {
+            action = _("reuse");
```
KKoukiou, line 93: reuse is essentially just doing the mount. IOW technically it's correct to write 'mount' there. I dont think we should explicitly mention reuse in the review page. Was this like this in the mockups? @garrett

src/components/storage/InstallationMethod.jsx
```diff
@@ -167,7 +191,9 @@ const usePageInit = () => {
     const { appliedPartitioning, partitioning } = useContext(StorageContext);
     // If the last partitioning applied was from the cockpit storage integration
     // we should not reset it, as this option does apply the partitioning onNext
-    const needsReset = partitioning.storageScenarioId !== "use-configured-storage" && appliedPartitioning;
+    const needsReset = partitioning.storageScenarioId !== "use-configured-storage" &&
+        partitioning.storageScenarioId !== "home-reuse" &&
```
KKoukiou, line 195: I think, all scenarios but for the 'use-configure-storage' should reset in the installation method page enter.

test/check-storage-home-reuse
```diff
@@ -0,0 +1,148 @@
...
+        s = Storage(b, m)
+        r = Review(b, m)
+
+        dev="vda"
+
+        i.open()
+        i.reach(i.steps.INSTALLATION_METHOD)
+        s.rescan_disks()
+
+        #s.wait_scenario_available("home-reuse", False)
```
KKoukiou, line 50: Pleave remove commented out code before this get's merged.

test/check-storage-home-reuse
```diff
@@ -0,0 +1,148 @@
...
+
+        s.set_partitioning("home-reuse")
+        i.reach(i.steps.REVIEW)
+
```
KKoukiou, line 54: No double empty lines

test/check-storage-home-reuse
```diff
@@ -0,0 +1,148 @@
...
+        r.check_disk_row(dev, parent=f"{dev}3", action="delete")
+        r.check_disk_row(dev, parent=f"{dev}1", action="delete")
+        r.check_disk_row(dev, "/boot", f"{dev}3", "1.07 GB", True, "xfs", is_encrypted=False)
+        r.check_disk_row(dev, "/", f"{dev}4", "12.8 GB", True, "btrfs", is_encrypted=False)
+        r.check_disk_row(dev, "/home", f"{dev}4", "12.8 GB", False, "btrfs", is_encrypted=False,
+                         action="reuse")
+
+        # TODO pixel test
+
+    @skipImage("btrfs support missing on fedora-eln image", "fedora-eln-boot")
```
KKoukiou, line 67: We dont really test on ELN anymore. You can just ignore it. We can also drop the rest of the decorator around ELN.

KKoukiou (commented):
This is looking great.

src/components/review/StorageReview.jsx
rvykydal: Agreed. Good point.

src/components/storage/InstallationMethod.jsx
rvykydal: As I write above, this is the only solution I was able to come so far with for https://github.com/rhinstaller/anaconda-webui/pull/424#issuecomment-2370817524 (the commit @rvykydal
[home reuse: FIXUP? do not reset home reuse after applying](https://github.com/rhinstaller/anaconda-webui/pull/424/commits/497b94b1a3040541969e9cda96794d429cd19a47))

src/apis/storage_partitioning.js
rvykydal: Then we need to know the scheme for the request - as can be seen below, the request differs for various schemes. I'll create an update reading it from the conf and passing the scheme to this function.

src/components/storage/InstallationMethod.jsx
KKoukiou: I think you need to change that to run this code only on component mount:
This diff would fix this I believe

``` diff
diff --git a/src/components/storage/InstallationMethod.jsx b/src/components/storage/InstallationMethod.jsx
index 617892cb..c6abca11 100644
--- a/src/components/storage/InstallationMethod.jsx
+++ b/src/components/storage/InstallationMethod.jsx
@@ -189,14 +189,16 @@ const InstallationMethodFooterHelper = () => {
 const usePageInit = () => {
     const { setIsFormDisabled } = useContext(FooterContext);
     const { appliedPartitioning, partitioning } = useContext(StorageContext);
+    const pageHasMounted = useRef(false);
+    // Always reset the partitioning when entering the installation destination page
     // If the last partitioning applied was from the cockpit storage integration
     // we should not reset it, as this option does apply the partitioning onNext
     const needsReset = partitioning.storageScenarioId !== "use-configured-storage" &&
-        partitioning.storageScenarioId !== "home-reuse" &&
-        appliedPartitioning;
+        appliedPartitioning &&
+        pageHasMounted.current === true;
 
     useEffect(() => {
-        // Always reset the partitioning when entering the installation destination page
+        pageHasMounted.current = true;
         if (needsReset) {
             resetPartitioning();
         } else {

```

rvykydal: > ```diff
> pageHasMounted.current === true;
> ```

It does (inverting this condition). Thank you!

src/apis/storage_partitioning.js
rvykydal: Fixed, all 4 schemes are supported internally now.

src/components/review/StorageReview.jsx
rvykydal: Updated.

test/check-storage-home-reuse
rvykydal: Removed

test/check-storage-home-reuse
rvykydal: Removed.

test/check-storage-home-reuse
rvykydal: I'll remove it in a separate PR.

rvykydal:
Seems that https://github.com/rhinstaller/anaconda-webui/pull/459 makes a few tests that succeeded yesterday fail. Going to update them.

rvykydal:
> Seems that #459 makes a few tests that succeeded yesterday fail. Going to update them.

Hm, the test results look a bit suspicous, re-running. Given we use updated Anaconda I wonder if there is some regression in Anaconda or Blivet.

rvykydal:
Updated the UI for the mockups https://issues.redhat.com/browse/INSTALLER-3479?focusedId=25032389&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25032389
Screenshot from 2024-10-14 11-05-18

KKoukiou (approved):

PR #431, opened 2024-09-05
reclaim: Adjust layout and alignment

garrett:
Fix up sizing and alignment to resemble the mockups more

### DIFF
--- a/src/components/storage/ReclaimSpaceModal.scss
+++ b/src/components/storage/ReclaimSpaceModal.scss
@@ -1,18 +1,11 @@
+@import "global-variables";
+
 #reclaim-space-modal-table {
   tr.reclaim-space-modal-table-row.reclaim-space-modal-device-level-1,
   tr.reclaim-space-modal-table-row.reclaim-space-modal-device-level-2,
   tr.reclaim-space-modal-table-row.reclaim-space-modal-device-level-3 {
     td {
-      padding-top: var(--pf-v5-global--spacer--sm);
-      padding-bottom: var(--pf-v5-global--spacer--sm);
-    }
-  }
-
-  @for $i from 1 through 10 {
-    tr.reclaim-space-modal-table-row.reclaim-space-modal-device-level-#{$i} {
-      td:nth-child(2) {
-        padding-inline-start: calc(var(--pf-v5-c-table--cell--PaddingLeft) + #{$i - 1} * var(--pf-v5-global--spacer--md));
-      }
+      padding-block: var(--pf-v5-global--spacer--sm);
     }
   }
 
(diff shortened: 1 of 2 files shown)


garrett:
I'm not quite done with this, but I wanted to share where I'm at currently, and make sure that it's not just on my computer over the weekend.

We might want to remove the flex for the actions too, as that affects the height of the rows and shouldn't be needed. This should simplify the JSX and CSS a little more. And I should look over the CSS again to see what's really needed or not (we might not need that subgrid, for example).

garrett:
Before:

image


After:

image

garrett:
It improves small sizes too.

Here's mobile before:

image

And mobile after:

image

src/components/storage/ReclaimSpaceModal.scss
```diff
@@ -33,6 +26,102 @@
...
+        var(--indent) * var(--pf-v5-global--spacer--sm)
+      );
+    }
+
+    // Remove labels with no content on small sizes
+    td:empty {
+      display: none;
+    }
+
+    tr {
```
KKoukiou, line 68: Isn't this too invasive on the Patternfly table structure? 
What's the problem with the default? 
Please add a comment if keeping this.

garrett: This isn't a PatternFly table; it's a Cockpit table which inherits from PF. As a result, we can't adjust it with PatternFly options.

But even still, if it were a plain PatternFly table: PatternFly does not officially support right aligning columns, which is completely absurd, as it's a basic thing that's been around for literal centuries in tables outside of computing and inside of computing basically everywhere else for decades. Some things, such as numbers, especially with units, really _need_ that right alignment. As do the actions at the end.

Additionally, I was testing for smaller widths, not just for mobile mode. As you can see the "before" screenshot has issues in smaller sizes. Smaller sizes show up when installing inside a VM... for example, in Cockpit Machines.

garrett: There are books about designing tables and layout (and I've read a few, such as _The Visual Display of Quantitative Information_ by Edward Tufte).

But here's a distilled version:
https://www.csescienceeditor.org/article/best-practices-in-table-design/

An even simpler version is this rule of thumb, adapted for computer UI:
1. Right align numbers (ideally separator align if possible, but that's trickier on the web)
2. Left align names (in LTR; this flips in RTL)
3. Use the same alignment for the headings
4. Use monospaced numbers (which can either be monospace font or tabular numbers in a proportional font) 

Sadly, PatternFly doesn't do any of this by default, except defaulting to left alignment.

KKoukiou (changes requested):
Please fix stylelint issues.

KKoukiou:
Will merge as part of https://github.com/rhinstaller/anaconda-webui/pull/491

PR #435, opened 2024-09-09
storage: use `console.warn` not `console.warning` to fix crash

AdamWill:
f66f8f4 mistakenly tried to log a warning with `console.warning`, which doesn't exist. It's `console.warn`. This means we still crash on the path where we crashed before, now with an error about `console.warning` being undefined.

### DIFF
--- a/src/actions/storage-actions.js
+++ b/src/actions/storage-actions.js
@@ -79,7 +79,7 @@ export const getDevicesAction = () => {
                             partName: partName.trim(),
                         };
                     } catch (e) {
-                        console.warning("Failed to get partition label", e);
+                        console.warn("Failed to get partition label", e);
                         devData.misc = {
                             fsLabel: "",
                             partName: "",


AdamWill:
btw, this also looks like an odd invocation? Per https://developer.mozilla.org/en-US/docs/Web/API/console/warn_static it seems like if you call it with a string followed by some other args, the subsequent args should be substitutions for the string...but this string doesn't contain anything that would be substituted by `e`. Am I reading the doc wrong, or was there perhaps meant to be a template in the string somewhere?

KKoukiou:
> btw, this also looks like an odd invocation? Per https://developer.mozilla.org/en-US/docs/Web/API/console/warn_static it seems like if you call it with a string followed by some other args, the subsequent args should be substitutions for the string...but this string doesn't contain anything that would be substituted by `e`. Am I reading the doc wrong, or was there perhaps meant to be a template in the string somewhere?

It will just be printed extra. See:
```
 If there are more substitution values than there are substitution strings, the extra values are themselves written to the console after the detailed assertion message in the same manner as when there's no format string.
```
In the  link you sent.

KKoukiou (approved):
Thank you Adam

PR #439, opened 2024-09-12
Don't merge

KKoukiou:


KKoukiou:
Verified that it fixes rhbz#2308279

PR #444, opened 2024-09-19
storage: move 'Modify storage' to the header kebab

KKoukiou:
Merge CockpitStorageIntegration and ModifyStorage files into one, having the "Modify storage" button split from the cockpit integration page implementation was slightly weird.

Also stop utilizing the 'isFormDisabled' in the
CockpitStorageIntegration, we don't disable the form anyhow from within cockpit-storage mode, so this was causing unnecessary complexity in the code.

Related to: https://issues.redhat.com/browse/INSTALLER-3972


KKoukiou:
> @KKoukiou I've played with the patch in my local VM and hit an issue when I
> 
>     * have 2 disks
> 
>     * create 1GB /boot 100MB biosboot, rest / in Storage Editor
> 
>     * then go back to installation
> 
>     * go to mount point assignment
> 
>     * the UI is blinking in some update loop perhaps
>       [webui_2disks_mpa_flashing.webm](https://github.com/user-attachments/assets/d0195e1e-3ef3-49c1-9de0-0ac8f8a60ac8)
>       May be worth trying to reproduce on your side.

Not sure if this bug is related to this PR (probably not) but I found what the issue was and sent two commits on top.

PR #451, opened 2024-09-25
Add anaconda-tui package to updates-image

jkonecny12:
We don't need other packages because these only have dependencies or they have GTK UI. However, we do want to keep tui code even when Web UI is the only requested option as the fallback solution.


jkonecny12:
Seems that this is fixing issue for @adamkankovsky PR https://github.com/rhinstaller/anaconda/pull/5739 but it is not enough for mine https://github.com/rhinstaller/anaconda/pull/5829.

Still we should merge this when it has a review.

KKoukiou (approved):

PR #456, opened 2024-09-30
Hotfix for Anaconda Wayland migration changes

jkonecny12:
There is a new patchset to migrate Anaconda to Wayland native application. This patchset is unfortunately quite big and complicated. https://github.com/rhinstaller/anaconda/pull/5829

Current Web UI testing depends on updates image creation from the Anaconda RPMs which are downloaded from Rawhide and/or COPR repos. However, we are hitting issues for updates images in this case.

- Removal of spoke in Anaconda will not be covered by updates image. This is expected because RPM don't have information that some file was removed. Usually this is not an issue because the removed code is not imported. Unfortunately, here the code is imported by our dynamic spoke importer. That will result in traceback because of missing imports in the removed spoke.
- The patchset in Anaconda is adding new dependencies required to be able to start the Wayland environment, however, these are not covered when an updates image is created because we can't tell which deps are new and should be part of the ISO.

We should resolve this somehow by improving the workflow in the future. For now we can continue with this hotfix.

### DIFF
--- a/scripts/makeupdates
+++ b/scripts/makeupdates
@@ -36,6 +36,16 @@ def addRpms(updates, add_rpms):
 
 def createUpdatesImage(cwd, updates):
     os.chdir(updates)
+    # FIXME: hotfix for issue with missing file - remove when new Wayland native
+    #        Anaconda is on Fedora Rawhide
+    if not os.path.exists("usr/lib64/python3.13/site-packages/pyanaconda/ui/tui/spokes/askvnc.py"):
+        try:
+            os.makedirs("usr/lib64/python3.13/site-packages/pyanaconda/ui/tui/spokes")
+        except FileExistsError:
+            pass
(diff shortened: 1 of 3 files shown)


KKoukiou (commented):
@jkonecny12 I see in the journals that the is anaconda traceback.

pyanaconda.errors.ExitError: /usr/libexec/anaconda/run-in-new-session exited on signal 9

https://cockpit-logs.us-east-1.linodeobjects.com/pull-456-a9f949e4-20240930-110347-fedora-rawhide-boot-anaconda-pr-5829-rhinstaller-anaconda-webui/TestBasic-testAboutModal-fedora-rawhide-boot-127.0.0.2-2201-FAIL.log.gz

jkonecny12:
We agreed to just go with the merge of Wayland migration and try to do the release ASAP. This has quite an issues so closing.

PR #461, opened 2024-10-04
build: spec: require anaconda-core 42.5 only on fedora 42

KKoukiou:
Fedora 41 does not contain that version, and also we don't plan to rebase there anymore, so it does not really matter.

Also clean up the specfile from conditions for Fedora < 41.


elkoniu (approved):
LGTM

PR #462, opened 2024-10-04
tests: apply updates.img with mkksiso script instead of inst.updates

KKoukiou:
Blocked on:

- [x] : https://github.com/cockpit-project/cockpituous/pull/623


jkonecny12:
Maybe I'm wrong but I think you will face an issue that running mkksiso in the container requires loop device available. So you need to have it created somewhere manually because of the issue with losetup.

PR #463, opened 2024-10-04
Add interactive-defaults.ks to updates image

jkonecny12:
The interactive-defaults is must have to enable testing of web ui on the boot.iso because we don't have support for package based installations there. This way we are serving the tarball for the VM which will be used as payload to overcome this missing implementation.

Do not SSH the interactive-defaults.ks into the VM when it's running but instead add the interactive-defaults.ks to the updates.img. The SSH copy of interactive-defaults.ks works but not reliably. The issue is that we are copying the interactive-defaults.ks to the running installation and if the file will land too late it won't be taken by the installer.

To avoid this issue add the interactive-defaults.ks content to the updates image before the VM is started.


jkonecny12:
@KKoukiou I'm not able to tell from the logs why I'm getting failures. Are you able to find more information?

test/machine_install.py
```diff
@@ -83,10 +84,15 @@ def _serve_payload(self):
...
+    def _write_interactive_defaults_ks(self, updates_image):
         payload_cached_name, http_payload_port = self._serve_payload()
         content = f'liveimg --url="http://10.0.2.2:{http_payload_port}/{payload_cached_name}"'
-        Machine.execute(self, f'echo \'{content}\' > /usr/share/anaconda/interactive-defaults.ks')
+        defaults_path = "usr/share/anaconda/"
+        with TemporaryDirectory() as tmp_dir:
+            os.makedirs(f"{tmp_dir}/{defaults_path}")
+            with open(f"{tmp_dir}/{defaults_path}/interactive-defaults.ks", "wt", encoding="utf-8") as f:
+                f.write(content)
+            os.system(f"cd {tmp_dir} && find . | cpio -c -o | gzip -9cv > {updates_image}")
```
KKoukiou, line 95: This does not extend the existing updates.img from what I understand - it overwrites it. Right?

jkonecny12: Oh, my true.... how could I do such a dumb mistake... I'll fix that, thanks for spotting this.

jkonecny12: Fixed!

jkonecny12:
I'm testing out switch from `gzip` to `pigz` which is used by Anaconda when updates images are created. This will hopefully help with
```
gzip: stdin: unexpected end of file
cpio: premature end of file
```
or
```
gzip: stdin: decompression OK, trailing garbage ignored
cat: write error: Broken pipe
```

and similar issues.

jkonecny12:
Thanks for the update @KKoukiou

jkonecny12:
I removed the WIP commits.

KKoukiou (approved):

PR #475, opened 2024-10-15
tests: do not update the anaconda packages till the wayland work reaches rawhide

KKoukiou:
Wayland changes bring many new dependencies, that if installed through updates.img, they cause multiple issues.

Work with the anaconda backend that already reached rawhide temporarily.


jkonecny12 (approved):
Great workaround.

PR #477, opened 2024-10-15
UX fixups followup to SPUR session with the team

KKoukiou:


src/components/storage/HelpAutopartOptions.jsx
```diff
@@ -4,8 +4,8 @@ const _ = cockpit.gettext;
 
 export const helpEraseAll = _("Remove all partitions on the selected devices, including existing operating systems. Make sure you have backed up your data.");
 
-export const helpUseFreeSpace = _("Keep current disk layout and only install into available space.");
+export const helpUseFreeSpace = _("Keep current disk layout and use available space, to dual-boot with another OS.");
 
-export const helpMountPointMapping = _("Manually select storage locations for installation.");
+export const helpMountPointMapping = _("Select existing partitions as mount points. Useful for existing custom layouts.");
```
garrett, line 9: Oh, that's based on an earlier version (up until this week :wink:). We were talking about revising it further in the chat, using select as the main title and assign as the sub-text (as to not be repetitive, as select is in the title and use is a different title for using the full disk).

I had "_Assign partitions to mount points. Intended for reusing an existing layout or after using a custom partitioner._" but I think we can make it shorter and still be fine enough... what do you think of something like this?

🔘 **Select existing partitions**
Assign partitions to mount points. Useful for pre-configured custom layouts.

(This removes the duplication of "select" and "existing" too.)

```suggestion
export const helpMountPointMapping = _("Assign partitions to mount points. Useful for pre-configured custom layouts.");
```

garrett (commented):
Currently working on building the VM. I'm looking at the diff right now and it looks good. I have a comment with a suggestion... what do you think?

src/components/storage/HelpAutopartOptions.jsx
garrett: (Also, I don't see the map mount points string changed in the diff, but perhaps I'm missing it?)

garrett:
I've added a disk to a test VM and it has mount point mapping with a bold string.

image

1. Mount point mapping should be called "Select existing partitions"
2. We shouldn't have bold text like that on a disabled entry, as it calls more attention to itself than other entries, even enabled ones.
3. When there are no partitions to map to mount points, we shouldn't show the option.

(Most options shouldn't be shown if they cannot be done.)

For example, here's the example mockup where there's just 1 empty disk with no partitions:

image

garrett:
We should remove "Operating system" and the space between it and Language, as we did run into some problems with verticality (especially with a complex storage and/or in small screen sizes, like VMs) and also since it mainly redundantly repeats what's in the header anyway.

image

garrett (approved):
I'm fine with what's here, and we can iterate with more changes on top. It's already a definite improvement. Thanks!

PR #491, opened 2024-10-29
Super picky UI review fixups round #2

KKoukiou:


garrett:
Dual boot and partition mapping options should not show up if there aren't partitions to work with. Only "Use whole disk" should always show up. (In the case of no disk, we could even have some kind of simple empty state, but it's hopefully such an uncommon case that having a disabled "Use whole disk" option is probably fine, really).

image

And when I add the disk, we shouldn't have the partition mapping:

image

The string "No usable devices on the selected disks." shouldn't be dark or bold, and the text doesn't make sense to have devices on the disks; it should be partitions. However, since this option shouldn't show up anyway, people probably shouldn't see this.

garrett:
## Encryption page

I think we need to change the sidebar?

And it'd probably be a good idea to migrate to the new design of the storage, with encryption as one of the settings, so things won't change too much in the future.

### Screenshot

image

### Mockup, eventual

Eventually, something like this:

Wizard, encryption(1)

(Text is super rough, just to convey the ideas. It needs more work. But those features would be coming later, so we'd have time. The point of this mockup is to be "aspirational" / "forward looking" / "defining a direction" / etc.)

### Mockup, encryption only

But pruned back to this, for the features we have now (encryption):

Wizard, encryption(3)

---

Changes between the screenshot and the pruned-down:

1. Side changes. (Do we have more than 1 sub-item right now?)
2. Has a main heading change.
3. Encryption heading changes.
4. Heading text changes.
5. Description text changes.

(The checkbox text stays the same.)

The description text is currently:
> Encryption helps secure your data, to prevent others from accessing it.
>
> Only new partitions will be encrypted. Existing partitions will remain untouched.

I'm suggesting:
> Secure your data using disk-based encryption. Only applies to new partitions.

as this removes a bit of redundancy, as only affecting new partitions implies that this won't affect non-new (existing) partitions.

I guess we could merge the two and go with
> Secure your data using disk-based encryption, to prevent others from accessing it. Only applies to new partitions.

although I think people would understand securing it will prevent others already? We could, then, rework it a little more to something like:
> Disk-based encryption prevents others from accessing your data. Only applies to new partitions.

M4rtinK (approved):

PR #494, opened 2024-10-30
Move disk encryption under generic storage configuration step

KKoukiou:
The motivation behind this was to prepare the ground to introduce more configuration options, like `swap` option.


garrett (approved):
Looks great, thanks for implementing it and being so quick with this!

PR #496, opened 2024-10-31
review: show device type if it's formatted as btrfs

KKoukiou:
So that the user get's informed for subvolumes.


M4rtinK (approved):
Looks good to me. :)

PR #497, opened 2024-11-01
Do not permit clicking 'Modify storage' button when installation has been started

KKoukiou:


src/components/app.jsx
```diff
@@ -51,7 +53,15 @@ export const ApplicationLoading = () => (
 export const Application = ({ conf, dispatch, isFetching, onCritFail, osRelease, reportLinkURL }) => {
     const [storeInitialized, setStoreInitialized] = useState(false);
     const [showStorage, setShowStorage] = useState(false);
+    const [currentStepId, setCurrentStepId] = useState();
     const address = useAddress();
+    const { path } = usePageLocation();
+
+    useEffect(() => {
+        if (path.length === 0 || path[0] !== currentStepId) {
+            setCurrentStepId(path[0]);
```
rvykydal, line 62: It is expected that in case of empty path currentStepId is set to undefined I think.

rvykydal (approved):
Looks good to me.

src/components/app.jsx
KKoukiou: It works yes, but I realize,maybe it would be cleaner to just omit it though. In that case path[0] !== currentStepId would also work, if path[0] is undefined.

PR #501, opened 2024-11-04
storage: cockpit: fix UI flickering when exiting cockpit storage

KKoukiou:
Previously we were resetting the checkStep prematurely, therefore causing a UI flicker between success and error mode in the cockpit storage integration exiting dialog.

Moving the checkStep reset before the 'prepare partitioning' step solves this issue.


adamkankovsky (approved):
LGTM, thanks.

PR #509, opened 2024-11-12
Disable remove extended partition for EFI operation systems

adamkankovsky:
image

### DIFF
--- a/src/components/storage/ReclaimSpaceModal.jsx
+++ b/src/components/storage/ReclaimSpaceModal.jsx
@@ -36,7 +36,14 @@ import {
     Text,
     TextContent,
 } from "@patternfly/react-core";
-import { CompressArrowsAltIcon, HddIcon, LockIcon, TrashIcon, UndoIcon } from "@patternfly/react-icons";
+import {
+    CompressArrowsAltIcon,
+    HddIcon,
+    LockIcon,
+    OutlinedQuestionCircleIcon,
+    TrashIcon,
+    UndoIcon
+} from "@patternfly/react-icons";
 
 import { isDeviceShrinkable, removeDevice, shrinkDevice } from "../../apis/storage_partitioning_automatic_resizable.js";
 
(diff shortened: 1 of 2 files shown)


garrett:
So we don't have an assiciation between extended and logical partitions, so we _can't_ do something like this:

image

So, instead of a top-down approach where you're trying to delete the container and the children, if you delete all the children, then the container disappears (in a bottom-up approach). The fact that the extended partition even exists is an artifact of a hack to shoehorn flexibility to allow logical partitions to exist next to primary partitions. Extended partitions don't matter overall; they're implementation details that exist to make logical partitions possible.

In other words, if we detect extended partitions (like on an MBR disk, with an empty "type"), we can label them and have a clickable info icon that can explain it, like this:

image

This is a quick mockup, so ignore some of the incorrect details (like missing resizes, having some actions disabled, etc.)

What you should notice:

- "extended partition" that is styled (grey and italics here)
- (?) icon that you can click
- popover with text that explains the situation
- no trash icon (not even a disabled one) as it's not possible to even do anything with this partition at all

The extended partition is kind of a "ghost" that Anaconda automatically deletes when it's not needed.

The text:

vda4 is an MBR extended partition that contains logical partitions. Removing all logical partitions will also automatically remove the extended partition.

"vda4" should be the partition name, and it should be bold.

garrett:
We chatted more, and the size is kind of redundant and mostly misleading on extended partitions too, as the logical partitions are the ones actually using size. So we should probably remove that from the list as well.

image

src/components/storage/ReclaimSpaceModal.jsx
```diff
@@ -359,14 +355,16 @@ const DeviceActionDelete = ({ device, hasBeenRemoved, newDeviceSize, onAction })
 
     // Disable the remove action for disk devices without partitions
     const isRemoveDisabled = device.type.v === "disk" && device.children.v.length === 0;
+    // Disable the remove action for partitions without children witch is not leaf node
```
KKoukiou, line 358: which are not lead nodes

src/components/storage/ReclaimSpaceModal.jsx
```diff
@@ -359,14 +355,16 @@ const DeviceActionDelete = ({ device, hasBeenRemoved, newDeviceSize, onAction })
 
     // Disable the remove action for disk devices without partitions
     const isRemoveDisabled = device.type.v === "disk" && device.children.v.length === 0;
```
KKoukiou, line 357: Please integrate the second check here:
 const isDiskWithoutChildren = device.type.v === "disk" && device.children.v.length === 0;
 const isExtendedPartition = device.attrs?.v.isleaf === "False" && device.children.v.length === 0;
 const isRemoveDisabled = isDiskWithoutChildren || isExtendedPartition;

KKoukiou (commented):
Can you explain in the commit message a bit more clearly why this is needed?

vojtechtrefny:
> So, instead of a top-down approach where you're trying to delete the container and the children, if you delete all the children, then the container disappears (in a bottom-up approach). The fact that the extended partition even exists is an artifact of a hack to shoehorn flexibility to allow logical partitions to exist next to primary partitions. Extended partitions don't matter overall; they're implementation details that exist to make logical partitions possible.

I'd add that in blivet in general extended partitions are treated as implementation detail, something that allows you to create more than 4 partitions on MSDOS partition table. In general we create and remove extended partitions automatically as needed. (It is possible to control this manually and blivet-gui allows that, but Anaconda itself, including manual partitioning does not.)
We also don't treat extended partitions as a containers -- the logical partitions are not children of the extended partition, all partitions are children of the disk. Extended partitions are a big hack and I am really glad that GPT is now default everywhere.

garrett (approved):
Thanks! :tada: 

Screenshot looks good to me (and very close to the mockup).

I'm happy with this, as long as @KKoukiou is happy.

KKoukiou (commented):
Looks good, can you please rebase and let the tests rerun?

KKoukiou (approved):
Would be really nice if you can test this one. Fine to do as followup.

PR #512, opened 2024-11-13
tests: add UEFI boot tests for dual booting with ubuntu

KKoukiou:


test/check-existing-system
```diff
@@ -67,6 +67,15 @@ class TestFedoraPlansUseFreeSpace(VirtInstallMachineCase):
         with b.wait_timeout(300):
             p.wait_done()
 
+        # FIXME: https://bugzilla.redhat.com/show_bug.cgi?id=2325707
+        # This should be removed from the test
+        if self.efi:
```
elkoniu, line 72: Where `self.efi` is initially defined in this base class?

test/check-existing-system
```diff
@@ -153,6 +166,12 @@ class TestFedoraPlansUseFreeSpace(VirtInstallMachineCase):
         self.install(needs_confirmation=True)
         self.verifyDualBoot(root_one_size=14.2, root_two_size=4.7)
 
+class TestFedoraPlansUseFreeSpaceBIOS(_TestFedoraPlansUseFreeSpace):
+    efi = False
```
elkoniu, line 170: If this should not be something like `self.efi`?

M4rtinK (approved):
Looks good as far as I can tell + one question. :)

test/check-existing-system
```diff
@@ -44,7 +44,7 @@ class TestExistingSystemFedora(VirtInstallMachineCase):
 
         r.check_deleted_system("Fedora Linux")
 
-class TestFedoraPlansUseFreeSpace(VirtInstallMachineCase):
+class _TestFedoraPlansUseFreeSpace(VirtInstallMachineCase):
```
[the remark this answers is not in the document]
KKoukiou: Correct. I think the test discovery tool picks up everything starting with Test / test

I do use this pattern in some places more, see:
test/check-basic:    def _testLogReview(self, b, m, idPrefix):
test/check-storage-mountpoints:    def _testEncryptedUnlock(self, b, m):


I dont object that's weird, I can think of a new convention and change it also for the existing usages.

Sometihing like: CommonTestFedoraPlansUseFreeSpace

test/check-existing-system
KKoukiou: Not sure what you ask

KKoukiou: Oh I guess you mean that I can instead setup it in the constructor with self.efi = False. Sure.

PR #520, opened 2024-11-18
dependabot: block Patternfly 6 till we are ready and implement few extra improvements on the config

KKoukiou:


jkonecny12 (approved):
LGTM!

PR #523, opened 2024-11-19
Add luks to storage context

adamkankovsky:


src/actions/storage-actions.js
```diff
@@ -152,3 +152,18 @@ export const setStorageScenarioAction = (scenario) => {
         type: "SET_STORAGE_SCENARIO",
     };
 };
+
+export const setLuksEncryptionAction = (isEncrypted) => ({
```
KKoukiou, line 156: I think three actions for this is overkill. (too much boilerplate code)
I would suggest to have one action setLuksEncryptionDataAction({ isEncrypted, passphrase, confirmPassphrase }) and make it possible to profide only part of the keys.
If a key is not passed the reducer should not overwrite the existing value. 

What do you think?

adamkankovsky: It would probably be too messy. I have modified the code to only one function.

KKoukiou (approved):
If you can introduce (possibly in followup) a check for this. Extend the existing testEncryptionPreserved

PR #524, opened 2024-11-19
Retheme to better match Fedora

garrett:
- Shift PF blue palette colors to use official Fedora blue and PF-inspired variants that better match Fedora
- Recolor Fedora logo to be the white-on-dark version (which isn't provided, so we filter it)
- Optimize link and buttons colors for both brand and legibility


garrett:
I think this should be nearly done, but it's EOBD, so I'm sharing what I have now as draft. I'll double check it tomorrow and may make some (possibly) smaller changes if it's not quite ready.

src/branding/fedora.scss
```diff
@@ -0,0 +1,83 @@
...
+    --pf-v5-c-content--Color: var(--_text);
+  }
+
+  html.pf-v5-theme-dark & {
+    background: var(--pf-v5-global--palette--blue-400);
+  }
+}
+
+/* Have the kebab toggle inherit the text color */
+#toggle-kebab {
```
KKoukiou, line 62: Is this section specific to the fedora colors?  Should it not be always black?  Can you show me some example?

src/branding/fedora.scss
```diff
@@ -0,0 +1,83 @@
...
+/* Special-case some light-mode optimizations */
+html:not(.pf-v5-theme-dark) {
+  /* The nav icon should be a similar blue to buttons and links */
+  .pf-v5-c-wizard__nav-link.pf-m-current::before {
+	--pf-v5-c-wizard__nav-link--m-current--before--BackgroundColor: var(--pf-v5-global--primary-color--100);
+  }
+
+  /* Give menu list items a blue tint in light mode */
+  .pf-v5-c-menu__list-item:has(.pf-m-selected) {
+    --pf-v5-c-menu__list-item--hover--BackgroundColor: var(--pf-v5-global--palette--blue-50);
```
KKoukiou, line 81: This is light grey normally right?
I dont object changing this really, but this goes beyond branding if we are changin grey to blue. It's really custom styling divergent from Patternfly's and should be isolated as override with our common way of overriding patternlfy CSS.

PR #536, opened 2024-11-28
Improve the language screen from new design

adamkankovsky:
Screen Shot 2024-12-09 at 20 06 36

### DIFF
--- a/src/components/localization/InstallationLanguage.jsx
+++ b/src/components/localization/InstallationLanguage.jsx
@@ -17,10 +17,9 @@
 
 import cockpit from "cockpit";
 
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect } from "react";
 import {
-    Alert,
-    Divider,
+    Button,
     Form,
     FormGroup,
     Menu,
@@ -28,10 +27,12 @@ import {
     MenuGroup,
     MenuItem,
     MenuList,
-    MenuSearch,
-    MenuSearchInput,
-    SearchInput, Title,
+    TextInputGroup,
+    TextInputGroupMain,
+    TextInputGroupUtilities,
+    Title,
 } from "@patternfly/react-core";
+import { SearchIcon, TimesIcon } from "@patternfly/react-icons";
 
 import { setLocale } from "../../apis/boss.js";
 import {
@@ -51,7 +52,6 @@ import "./InstallationLanguage.scss";
 const _ = cockpit.gettext;
 
 const getLanguageEnglishName = lang => lang["english-name"].v;
-const getLanguageId = lang => lang["language-id"].v;
 const getLanguageNativeName = lang => lang["native-name"].v;
 const getLocaleId = locale => locale["locale-id"].v;
 const getLocaleNativeName = locale => locale["native-name"].v;
@@ -62,8 +62,8 @@ class LanguageSelector extends React.Component {
         this.state = {
             search: "",
         };
+        this.initiallySelectedLanguage = props.language;
 
-        this.updateNativeName = this.updateNativeName.bind(this);
         this.renderOptions = this.renderOptions.bind(this);
     }
 
@@ -80,19 +80,13 @@ class LanguageSelector extends React.Component {
         }
     }
 
-    async updateNativeName (localeItem) {
-        this.props.setNativeName(getLocaleNativeName(localeItem));
-    }
-
     renderOptions (filter) {
         const { commonLocales, languages } = this.props;
         const idPrefix = this.props.idPrefix;
-        const filterLow = filter.toLowerCase();
 
         const filtered = [];
+        const filterLow = filter.toLowerCase();
 
-        // Is set to true when the first instance of a selected item is found.
-        let foundSelected = false;
         // Returns a locale with a given code.
         const findLocaleWithId = (localeCode) => {
             for (const languageId in languages) {
@@ -106,13 +100,11 @@ class LanguageSelector extends React.Component {
             console.warn(`Locale with code ${localeCode} not found.`);
         };
 
-        // Returns a new instance of MenuItem from a given locale and with given prefix in it's key
-        // and id.
+        // Helper to create a menu item
         const createMenuItem = (locale, prefix) => {
             const isSelected = this.props.language === getLocaleId(locale);
-
             // Creating a ref that will be applied to the selected language and cause it to scroll into view.
-            const scrollRef = (isSelected && !foundSelected)
+            const scrollRef = (isSelected)
                 ? (ref) => {
                     if (ref) {
                         ref.scrollIntoView({ block: "center" });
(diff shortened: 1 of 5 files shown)


src/components/localization/InstallationLanguage.jsx
```diff
@@ -103,89 +106,107 @@ class LanguageSelector extends React.Component {
...
-                        }
+                        {suggestedItems}
                     </MenuGroup>
-                    <Divider />
                 </React.Fragment>
             );
         }
 
-        // List alphabetically.
-        const languagesIds = Object.keys(languages).sort();
```
KKoukiou, line 166: @garrett should alphabetical sorting persist or not?

src/components/localization/InstallationLanguage.jsx
```diff
@@ -103,89 +106,107 @@ class LanguageSelector extends React.Component {
...
+                        <div>
+                            <span>{getLocaleNativeName(locale)}</span>
+                            {isSelected && (
+                                <span id="selected-icon" aria-hidden="true">
+                                    <CheckIcon />
+                                </span>
+                            )}
+                        </div>
+                        <span id="item-translation">
+                            {getLanguageEnglishName(locale)}
```
KKoukiou, line 144: @garrett i think we need some clarification here.
What it the right column? IN the mockups it's inconsistent, sometimes translated locale to english, sometimes the native. Please clarify.

Also we would need some help to restrain the height of the menu, as it currently spans larger than the available space, and it creates the common issue with the double scrollbar.

KKoukiou (commented):
@garrett we need your input here.

garrett:
Hey, so... this isn't a "new" design. It was from a few years ago, as a rough draft if we couldn't get GNOME Initial Setup included... in Fedora 37? And GIS was considered a "done deal" ever since, until recently. I'm just sharing the context, as it was quick and unfinished then and it's dusty now, as it did get messed up during the Penpot 2.0 release (April) due to a bug where the text was reset.

The first is supposed to be English + the detected set of languages based on region (when there's network), with a fallback of the most common languages. The most likely would be at the top. For example, in Germany, it'd be "Deutsch" at the top with "English" shortly after. In European countries, there should probably be other European languages. People move around; a few in the Anaconda and Cockpit teams can attest to that. Someone might want a local language, whereas someone else might want their native language. (It partially depends on how finished a translation is in Anaconda and the rest of Fedora too. We might want to limit the suggest to things that are translated over a threshold as well, but as discussed a couple years ago, that could be tricky, as it's not just Anaconda, but the rest of Fedora, including whichever desktop is being installed)

The second set should be "Additional languages" alphabetically sorted. That's rather clear.

As for what should go in the side, we wanted to make this similar to what GNOME has, as that's been around for a while and it'll be familiar with what people will use on an installed system after installing Workstation.

That looks like this:

image

The left is the language, the right is the variation of the language. Both parts are in the native language so someone can read and select them.

When filtered to Deutsch (German), ait lists the variations per country:

image

It also matches for translated language names too:

image

And place names, even translated ones:

image

image

And even abbreviations, like CZ:

image

However, not all, as UK isn't in any of the strings, apparently:

image

Basically, it needs to match on strings that aren't in the UI, as long

garrett:
Things to note about what others do:

1. GNOME (desktop, not GIS) doesn't have the suggested languages, but only shows them in alphabetical order. 
   - But that's because the system is already installed and someone chose a language prior.
   - GNOME's list of languages is much shorter than ours by default. I have the unofficial "show extended input sources" on in GNOME Tweaks, which makes the list a LOT longer.
      image
   - GNOME Initial Setup has something similar, but puts probable suggest languages at the top (not grouped, not divided, just at the top with the rest alphabetical, and less common ones hidden with a ... entry — note: all show up when explicitly searched for, or when expanding)
     image


2. macOS has suggested languages at the top of their installer; and their OS has a divider whereas their installer does not. (So they have the grouping too, just not as explicit.)

   - Installer: image (from https://techlabs.blog/categories/guides/install-macos-on-windows-in-vmware-workstation-using-a-virtual-disk-image)
   
   - Post-install configuration (note how it even has different formatting, with translated languages too, oddly different from the installer): image (from https://www.flagsarenotlanguages.com/blog/)

3. Windows is a bit different. They have builds per region, as well as international buildes.
   
   - You get the pick of German or English here: 
      image (from https://superuser.com/questions/1058237/how-to-change-windows-10-installation-language)
      image (from https://www.ghacks.net/2021/09/18/how-to-change-the-windows-11-display-language/)

   - 

4. Android
   - I don't have a screenshot of the initial setup
   - After installation: They actually do two lists with headers, like my mockup:
      - image
      - (I have developer mode on, so there are strange debug languages included in the mix.)
    - The groups are "Suggested" and "All languages", but there aren't duplicates, so "All" really should be "Additional"; you can see this if I filter it down to Deutsch, for example:
         image

garrett:
(Working on writing a response summary, taking the above into consideration, with a revised mockup... it's coming soon)

garrett:
## Story time: American near Munich

American visiting the Red Hat office in Grasbrunn (EMEA office, outside of Munich) installs Fedora 42. Anaconda detects that it is in Germany, so it preselects German (Deutsch), but has some other suggested languages at the top, based on what's popular in the region and also has good language support in both Anaconda and Fedora.

Welcome, languge, for boot iso(5)

The suggested list has the preselected, likely option at the top. Others are sorted alphabetically. There's a list of everything else below.

This person clicks on the "English / United States" option and the UI changes from German to English; the order of the items do not change:

Welcome, languge, for boot iso(7)

Entries for languages have the language and the location, both in the native language.

---

## A different story: French Canadian person searching in the USA

A person from Canada who feels more comfortable in French is in the USA, and searches for "french", as they're using a US keyboard without support for the "ç" key.

Welcome, languge, for boot iso(8)

If it's too difficult to have the two groups when doing a search, we could collapse it all down to have it just one group and alphabetical for everything:

Welcome, languge, for boot iso(9)

KKoukiou:
@garrett @adamkankovsky for splitting the locale name from the territory name like in the mockups we would need backend changes as now the backend provides, `language name (territory name)` combined.
I think it's fine for first pass to not bother about keeping these split, as it's purely cosmetic improvement.

KKoukiou (approved):

KKoukiou (approved):

garrett:
- list is too tall in small sizes so that content scrolls off screen, yet too short when it's by itself on larger screens
   - we shouldn't have two scrollbars, the list shouldn't be quite as tall when there isn't room for it
   - the ideal fix would be to have the full layout considered, but that'd probably be tricky as patternfly nests elements in the DOM needlessly, causing a huge nested mess of divs, which makes it hard to work with
   - a "simpler" fix is to figure out a magic number (with units) that estimates the size of the heading, the footer, and then the approximate height from the top of the content to the element, and the rest of the page... yeah, not great... it's like 
   - heading: 84, footer: 44, content (about from header): 158, necessary padding underneath: 8px
   - this means it'd be something like `max-height: calc(100vh - 84px - 44px - 158px - 8px - 50px)` (not sure where 50px comes from, but I tried it live in devtools).... and we'd need more when there are more widgets... this is an improper hack; the proper solution would really need to be using a layout system like grid... (flex would even have some issues, but could probably be made to work too)... and this calc math wouldn't even work properly on larger sizes, which is why it really needs to be a flex (which only stretches at the children level) or ideally a grid (which can size based on the parent and the children, instead of mainly just the children)
     image

- there are duplicate entries, which causes issues (such as having two selected)
     image


- entries shift around; they shouldn't... they should stay in the same place as the initial rendering
  - items shouldn't be added to suggested

    image

- when going forward and back, the selected item should definitely be visible (it might need the scrollIntoView() function, probably with the options block: 'nearest' and behavior: 'instant')

- order isn't right: it should have the most likely suggestion first (the one that's selected by default), then the rest of suggested in alphabetical order, then it should have another group of additional languages that does not include the ones in the suggestions

- searching (filtering down) should either have a group of everything in alphabetical order _or_ keep groups with headings when there are items in a group (and don't show the groups when there are no matches in it)

- heading spacing was wrong, but I've already fixed those spacing related issues in a commit and pushed it

PR #542, opened 2024-12-04
tests: introduce run_boot decorator for boot mode-specific tests

KKoukiou:
See also https://github.com/cockpit-project/bots/pull/7182

### DIFF
--- a/test/run
+++ b/test/run
@@ -21,6 +21,7 @@ RE_EXPENSIVE='E2E'
 
 # every known case needs to set RUN_OPTS to something non-empty, so that we can check if we hit any branch
 case "${TEST_SCENARIO:=}" in
+    *efi*) RUN_OPTS="$ALL_TESTS"; export TEST_FIRMWARE="efi" ;;&
     *expensive*)
         RUN_OPTS="$(echo "$ALL_TESTS" | grep -E "$RE_EXPENSIVE")"
         ;;
--- a/test/README.rst
+++ b/test/README.rst
@@ -210,9 +210,17 @@ You can set these environment variables to configure the test suite::
     TEST_OS    The OS to run the tests in.  Currently supported values:
                   "fedora-rawhide-boot"
 
-    TEST_SCENARIO There are two supported non-default scenarios:
+    TEST_SCENARIO These are the supported scenarios:
                      `cockpit-pr-N`: to install the cockpit packages from its pull request #N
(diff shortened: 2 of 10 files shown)


martinpitt (commented):
Thanks! This looks good enough. Feels a bit weird to have both decorators and the Test..{EFI,Bios} class split -- feels like making the decorator work for classes would be a bit more elegant. But this is completely bikeshedding, and no need to waste time on that if you are happy with this.

[this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-542-b48973ed-20241205-112656-fedora-rawhide-boot-other-efi/log) might be relevant or just a flake? I'll retry it to compare.

KKoukiou:
> Thanks! This looks good enough. Feels a bit weird to have both decorators and the Test..{EFI,Bios} class split -- feels like making the decorator work for classes would be a bit more elegant. But this is completely bikeshedding, and no need to waste time on that if you are happy with this.
> 
> [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-542-b48973ed-20241205-112656-fedora-rawhide-boot-other-efi/log) might be relevant or just a flake? I'll retry it to compare.

Thhe Test..{EFI, BIOS} class split was eliminated in the second commit by the TEST_SCENARIO=efi :)

test/run
```diff
@@ -21,6 +21,7 @@ RE_EXPENSIVE='E2E'
 
 # every known case needs to set RUN_OPTS to something non-empty, so that we can check if we hit any branch
 case "${TEST_SCENARIO:=}" in
+    *efi*) RUN_OPTS="$ALL_TESTS"; export TEST_FIRMWARE="efi" ;;&
```
rvykydal, line 24: I am confused by the '&' at the end of the line.

rvykydal (approved):
Looks good to me, thank you.

test/run
martinpitt: It will make the `case` evaluation continue with the next one. I.e. the scenarios are not mutually exclusive, but combine.

KKoukiou: he ;& in a case statement in a Bash script allows execution to continue with the next pattern block in the case statement, regardless of whether the next pattern matches or not. 

The 'efi' test scenario it not mutually exclusive with the `expensive` and `other` scenarios, they can be used together.
So `expensive/efi`, `other/efi`. I will update the documentation in test/README.rst to reflect that.

PR #544, opened 2024-12-05
Firefox 133 fixes

garrett:
I've modified the theme in a few ways, at least for `live`; this could apply to `default` too, perhaps.

1. I fixed a parsing error. I _think_this works for that setting, but I'm not sure.
2. I reworked the way the UI is hidden when there's 1 tab.


garrett:
Moved the last commit to https://github.com/rhinstaller/anaconda-webui/pull/545 — this one only has the fixes now.

KKoukiou (approved):

PR #545, opened 2024-12-05
Allow Anaconda to override Firefox shortcuts

garrett:
This tells Firefox to allow us to override its built in shortcuts, so we could rewrite what ctrl-w, ctrl-q, ctrl-t, and the rest do from within Anaconda.

(It might or might not still work in current Firefox; the docs I saw about it are years old.)


KKoukiou (commented):
Lets also do the actual change of overriding shortcuts and merge then.

KKoukiou:
Please reopen if you plan to work on that, cleaning up the stale open PRs.

PR #551, opened 2024-12-13
Parse new 'hidden_screen' configuration file options

KKoukiou:


src/components/steps.js
```diff
@@ -26,5 +29,30 @@ export const getSteps = (...args) => {
...
+     * 2. Statically: Configured via the 'hidden_screens' key in the 'anaconda.conf'
+     *    For example, the 'Account Creation' screen is hidden in the 'Workstation' ISO
+     *    because this step is handled by the 'Gnome Initial Setup' tool during the first boot.
+     */
+    return stepsOrder
+            .filter(s => {
+                const isHidden = hiddenScreens.includes(s.id);
+
+                if (isHidden) {
+                    debug(
```
M4rtinK, line 45: This goes only to the browser console, right ? Not relevant for this PR, but I'm thinking we might want to maybe synchronize the explicit log messages from the Web UI to the Journal as well, via cockpit bridge in the future ? We use this quite heavily with the GTK UI, where we can see user visiting screens before a crash for example - or the reasons for a screen being hidden. Without this, debugging more complex Web UI issues where we only have the log dump available from a user/customer, might be hard to impossible. :P

src/components/installation/InstallationProgress.jsx
```diff
@@ -45,27 +45,27 @@ import "./InstallationProgress.scss";
 
 const _ = cockpit.gettext;
 const N_ = cockpit.noop;
-const idPrefix = "installation-progress";
+const idPrefix = "progress";
```
M4rtinK, line 48: I'm thinking that perhaps if this is now essentially a part of the stable API (as people might have their own config files via the drop-in dir) maybe it should actually have some prefix ?

Eq. `anaconda-screen-foo` or `anaconda-feature-foo`

That would also make it easier to find all places where it is used & possibly make it less likely to be broken by mistake. 

And lastly if we need to change such a string in the future (eq. splitting a screen or changing its purpose) while keeping backwards compatibility, it would be easier with a well-known prefixed id than just a generic name.

In any case, this is the best time to think about it, while we can be reasonably sure no one is using this yet :)

M4rtinK (approved):
Looking good as far as I can tell technically. :)

Added a few notes/suggestions about logging & potentially prefixing the screen/feature id strings.

src/components/installation/InstallationProgress.jsx
KKoukiou: I like the suggestion with anaconda-screen vs anaconda-feature prefixes. Now we support only hidding screens, maybe in the future we would need to hide features.

M4rtinK: Sounds good. :)

M4rtinK (approved):
Looking good, thanks! :)

PR #561, opened 2024-12-18
Adding keyboard settings with localectl

adamkankovsky:
image

### DIFF
--- a/src/actions/localization-actions.js
+++ b/src/actions/localization-actions.js
@@ -17,10 +17,12 @@
 
 import {
     getCommonLocales,
+    getCompositorSelectedLayout,
     getLanguage,
     getLanguageData,
     getLanguages,
     getLocaleData,
+    getLocaleKeyboardLayouts,
     getLocales,
 } from "../apis/localization.js";
 
@@ -66,6 +68,18 @@ export const getCommonLocalesAction = () => {
         return dispatch({
             payload: { commonLocales },
             type: "GET_COMMON_LOCALES"
+        });
+    };
+};
+
+export const getKeyboardLayoutsAction = ({ language }) => {
+    return async (dispatch) => {
+        const keyboardLayouts = await getLocaleKeyboardLayouts({ locale: language });
+        const compositorSelectedLayout = await getCompositorSelectedLayout();
+
+        dispatch({
+            payload: { compositorSelectedLayout, keyboardLayouts },
+            type: "GET_KEYBOARD_LAYOUTS"
         });
     };
 };
--- a/src/apis/localization.js
+++ b/src/apis/localization.js
@@ -17,7 +17,7 @@
 
 import cockpit from "cockpit";
 
-import { getLanguageAction, getLanguagesAction } from "../actions/localization-actions.js";
+import { getKeyboardLayoutsAction, getLanguageAction, getLanguagesAction } from "../actions/localization-actions.js";
 
 import { debug } from "../helpers/log.js";
 import { _callClient, _getProperty, _setProperty } from "./helpers.js";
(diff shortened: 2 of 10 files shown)


src/actions/localization-actions.js
```diff
@@ -50,12 +51,16 @@ export const getLanguageDataAction = ({ language }) => {
 
 export const getLanguageAction = () => {
     return async (dispatch) => {
-        const language = await getLanguage();
-
-        return dispatch({
-            payload: { language },
-            type: "GET_LANGUAGE"
-        });
+        try {
```
KKoukiou, line 54: Please avoid using try catch here. Failures in these functions shoudl not be caught and ignored, but rather propagated to to main error handler, where they are raised as CriticalErrors and stopping the installation.

src/actions/localization-actions.js
```diff
@@ -50,12 +51,16 @@ export const getLanguageDataAction = ({ language }) => {
...
+            const language = await getLanguage();
+            dispatch({
+                payload: { language },
+                type: "GET_LANGUAGE"
+            });
+            dispatch(getKeyboardLayoutsAction({ language }));
```
KKoukiou, line 60: Please dont add this here. Find where getLanguageAction is called in src/apis/localization.js and do this bellow that.

src/actions/localization-actions.js
```diff
@@ -69,3 +74,17 @@ export const getCommonLocalesAction = () => {
...
+
+export const getKeyboardLayoutsAction = ({ language }) => {
+    return async (dispatch) => {
+        try {
+            const keyboardLayouts = await getKeyboardLayoutsForLanguage(language);
+            dispatch({
+                payload: { keyboardLayouts },
+                type: "GET_KEYBOARD_LAYOUTS"
+            });
+        } catch (error) {
```
KKoukiou, line 86: No try catch, this needs to be handled as CriticalError.

src/apis/localization.js
```diff
@@ -139,3 +139,109 @@ export const getLocaleData = ({ locale }) => {
...
+};
+
+/**
+ * @param {string} lang         Language id
+ * @returns {Promise}           Resolves a list of keyboard layouts for the specified language
+ *                              in format {description:, id:}
+ */
+export const getKeyboardLayoutsForLanguage = async (lang) => {
+    const normalizedLang = normalizeLanguage(lang);
+    const filePath = "/usr/share/X11/xkb/rules/evdev.lst";
```
KKoukiou, line 175: Parsing this file sounds like not the most robust solution. I dont object it, but maybe we could think of something more stable? 

This info could be retrieved I think from the python bindings of xdb, python-xkbcommon

So we could have a python script that generates this, converts it into a JSON file, which we can easily consume from Javascript. 

@jkonecny12 what do you think? Do we want this solution, as long as it's tested and wait for Gnome to resolve this by implementing the localed APIs for Workstation?

src/components/localization/InstallationLanguage.jsx
```diff
@@ -294,22 +297,66 @@ class LanguageSelector extends React.Component {
...
+
+    const selectedValue =
+        selectedKeyboard ||
+        (keyboards.length > 0 ? `${keyboards[0].layoutId}:${keyboards[0].variantId || ""}` : "");
+
+    return (
+        <FormSelect
+          value={selectedValue}
+          onChange={handleChange}
+          aria-label="Select Keyboard Layout"
```
KKoukiou, line 329: You dont need aria label, just add an id to the element, and supply that to fieldId of the FormGroup.

KKoukiou (commented):
Great job, please add tests.

jkonecny12 (changes requested):
We should use existing code in Anaconda. In general keyboard layouts logic is hard a lot of specifics which might give us a lot of headache especially if we would keep the code on two places.

src/components/localization/InstallationLanguage.jsx
```diff
@@ -295,21 +298,23 @@ class LanguageSelector extends React.Component {
...
     return (
         <>
-            <Title
-              headingLevel="h3"
-            >
-                {_("Choose a language")}
-            </Title>
-            <Form>
-                <FormGroup>
+            <Title headingLevel="h3">{_("Choose Language and Keyboard")}</Title>
```
KKoukiou, line 315: Adam can you also please remove this? It's not in the mockups.

KKoukiou:
> The new test is expected to fail for fedora-42. Please use skipImage for fedora-42-boot, testKeyboardLayoutSelection

Or actually sorry this will break 42 actually. Let's display the keyboard selection only if the API exists?

KKoukiou (changes requested):
Please check comments and rebase.

KKoukiou (approved):

adamkankovsky (commented):
It looks very good. Thank you. 
I'm just not sure if the pr includes test/static-code

KKoukiou:
Fedora-42-boot: # testBasic (__main__.TestStorageMountPoints.testBasic)  - Unnrelated flake.

PR #578, opened 2025-01-03
tests: prefer composition over inheritance

KKoukiou:


jkonecny12:
Can't help myself but seeing the first commit as faking the inheritance. I think we shouldn't use `__getattr__` if we want to go with this.

rvykydal:
I'd be also careful with using __getattr__. OTOH tempted to say it is just tests, but still.
Although it is very unlikely situation, the mechanism can't handle methods with the same in the components, right?

Would it be too bad to use components explicitly by clients?

PR #579, opened 2025-01-03
Stop importing storage helpers

KKoukiou:


jkonecny12 (approved):
LGTM! 

I think it is fine to expect the same disk name in the VM all the time.

rvykydal (approved):

PR #585, opened 2025-01-09
Fix scripts/makeupdates root detection

AdamWill:
It currently only works if you have run certain `make` targets - anaconda-webui.spec is not part of the version-controlled repo, it's generated. It seems safer to look for a file under version control.

I'm not sure the script is much use for anaconda-webui anyway - I don't think there's a mechanism to apply changes to the jsx files, for e.g. - but it doesn't hurt to improve this.

### DIFF
--- a/scripts/makeupdates
+++ b/scripts/makeupdates
@@ -50,7 +50,7 @@ def usage(cmd):
 def main(argv):
     prog = os.path.basename(argv[0])
     cwd = os.getcwd()
-    spec = os.path.realpath(cwd + '/anaconda-webui.spec')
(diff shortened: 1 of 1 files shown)


AdamWill:
ohh, hmm. so an `anaconda-webui.spec` *does* show up in the root of the repo...but only if you run some `make` targets, it's not under version control. Looking for a file that *is* under version control does seem better.

KKoukiou (approved):

KKoukiou:
Carefully ignoring the test failures, they are fixed already by https://github.com/rhinstaller/anaconda-webui/pull/589

PR #592, opened 2025-01-16
Test compose builds

KKoukiou:


Makefile
```diff
@@ -189,8 +189,19 @@ payload: bots
...
 .PHONY: images
-images: bots
-	bots/image-download $(TEST_OS) debian-stable ubuntu-stable fedora-41
+images: bots cloud-images iso-images
+
+.PHONY: cloud-images
+	bots/image-download debian-stable ubuntu-stable fedora-41
+
+.PHONY: iso-images
+	# If a compose if specified download from the compose otherwise
```
martinpitt, line 198: I'm not sure this syntax works. I've always just seen `PHONY: <name>` as a declaration, not starting a rule. I.e. more like

```
.PHONY: images cloud-images iso-images

cloud-images:
          <commands>
```

etc.

If that combined phony/rule declaration works, then fine of course and TIL!

But also, do you really need three different rules here? Just make `images` download the stuff that you need?

Makefile
```diff
@@ -189,8 +189,19 @@ payload: bots
...
+	# download the ISO from Cockpit image server
+	if [ -n "$TEST_COMPOSE" ]; then \
+		test/download-iso $TEST_OS $TEST_COMPOSE; \
+	else \
+		bots/image-download $TEST_OS; \
```
martinpitt, line 203: The dot will break the command.

test/machine_install.py
```diff
@@ -116,7 +116,13 @@ def start(self):
 
         self.http_updates_img_port = self._serve_updates_img()
 
-        iso_path = f"{os.getcwd()}/bots/images/{self.image}"
+        # If custom compose if specified for fetching the image then use that
+        # else get the image from the bots directory
+        compose = os.environ.get("TEST_COMPOSE")
+        if compose:
```
martinpitt, line 122: Completely optional: You can use a walrus for this:

```py
if compose := os.e.g(...):
    use compose here
```

test/vm.install
```diff
@@ -37,6 +37,14 @@ def vm_install(image, verbose, quick):
         # Make sure builder can build packages if required, /var/tmp/build needs to be owned by builder
         machine.execute("su builder -c 'mkdir -p /var/tmp/build/SRPMS'")
 
+        # If compose is specified enable the custom compose repository and disable the existing ones
+        if compose:
+            machine.execute("sudo sed -i 's/enabled=1/enabled=0/' /etc/yum.repos.d/*.repo")
```
martinpitt, line 42: No need for sudo, m.execute already runs as root.

martinpitt (changes requested):
Nice! This is useful even without any extra integration.

Makefile
KKoukiou: test/download-iso "$(TEST_OS)" "$(TEST_COMPOSE)"; \    

This should work, right?

test/machine_install.py
KKoukiou: Changed. thanks

Makefile
martinpitt: Err, which dot... I'm not crazy I  did see it. But oops, I just scratched off a speck of dirt from my monitor which happened to be on a very strategic place :see_no_evil: Sorry!

Makefile
```diff
@@ -190,7 +190,15 @@ payload: bots
 
 .PHONY: images
 images: bots
-	bots/image-download $(TEST_OS) debian-stable ubuntu-stable fedora-41
+	# Download cloud images
+	bots/image-download debian-stable ubuntu-stable fedora-41
```
martinpitt, line 194: Unrelated to this PR, but is that hardcoding of f-41 necessary? That'll just bitrot over time, no? And if you test rawhide or a compose, do you really need it?

test/download-iso
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -xeu
+
+TEST_OS=$1
+TEST_COMPOSE=$2
```
martinpitt, line 6: Please quote. Test composes shouldn't have spaces in them, but it's a good habit.

test/download-iso
```diff
@@ -0,0 +1,29 @@
...
+BASE_OS="${TEST_OS/\-boot/}"
+RELEASE="${BASE_OS/fedora-/}"
+
+# Check if file already exists
+if [ -f test/images/$TEST_COMPOSE.iso ]; then
```
martinpitt, line 11: quote please

martinpitt (approved):
Thanks! Good enough, just some little suggestions in case you want to update again.

jkonecny12 (approved):
Looks great to me thanks!

Makefile
KKoukiou: We are using fedora-41 image, debian-stable and ubuntu-stable images as a base images for the target disk of the installation , and then rawhide-iso for the installer media. 
So the user might want Fedora rawhide, after shrinking their pre-existing Fedora/Ubuntu/Debian installation.

I guess I can replace existing test to just use the Fedora rawhide image instead of the Fedora-*. It's just weird for the user to want to install Fedora-rawhide next to existing Fedora-rawhide (i think it might not even be supported by Anaconda currently)

martinpitt: It's fine, and I figure you'll learn when to move it soon enough once we'll drop fedora-41 (and forgot about this reference, as it doesn't appear in bots testmap). :grin:

PR #593, opened 2025-01-16
infra: introduce daily cron job to trigger tests against latest compose

KKoukiou:
Rebased on top of: https://github.com/rhinstaller/anaconda-webui/pull/592

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,21 @@
+name: nightly
+on:
+  schedule:
+    - cron: '0 1 * * *'
+  # can be run manually on https://github.com/rhinstaller/anaconda-webui/actions
+  workflow_dispatch:
+jobs:
+  trigger:
+    permissions:
+      statuses: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Trigger the latest compose scenario
+        run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  schedule:
+    - cron: '0 1 * * *'
+  # can be run manually on https://github.com/rhinstaller/anaconda-webui/actions
+  workflow_dispatch:
+jobs:
+  trigger:
+    permissions:
+      statuses: write
+    runs-on: ubuntu-22.04
```
martinpitt, line 11: Is there a specific reason for this? Unless there is, please use "ubuntu-latest". It's 24.04 now, and it'll just move with the time. There's nothing special going on here, just a little bit of Python.

martinpitt (changes requested):
Thanks! When rebasing, I suggest to fix `runs-on:`, otherwise this tests itself.

.github/workflows/nightly.yml
KKoukiou: Nope replaced.

martinpitt (approved):
I'd say ship it! There is relatively little that can go wrong here (yet), at worst you'll get an issue filed every night about a failing test. But that's not risky, and we can easily adjust the workflow. Moreover, once it's on main, you can test changes much more easily by running the workflow from an origin branch (this isn't possible yet, it must *exist* on main).

PR #594, opened 2025-01-17
tests: fix nightly action for testing latest compose

KKoukiou:
We want to test the compose untouched, that means anaconda-webui package that is inside the compose. That means we also need to checkout the anaconda-webui repository, before running the tests to the tag used in the compose.
This is problematic, because our support for testing composes did not yet land in rawhide, therefore no tag contains that currently.

Let's test against anaconda-webui from main branch, temporarily.


KKoukiou:
[fedora-rawhide-boot/compose-Fedora-Rawhide-20250117.n.0](https://cockpit-logs.us-east-1.linodeobjects.com/pull-594-44157a68-20250117-143149-fedora-rawhide-boot-compose-Fedora-Rawhide-20250117.n.0/log.html)
fedora-rawhide-boot/compose-Fedora-Rawhide-20250117.n.0 scenario passed

martinpitt (approved):
This all looks reasonable. Hopefully we can clean up the first commit soon, i.e. after the next release. Thanks!

PR #601, opened 2025-01-20
tests: report passed testcases to Fedora Wiki

KKoukiou:


test/anacondalib.py
```diff
@@ -273,9 +276,10 @@ def tearDown(self):
         super().tearDown()
 
 
-def test_plan(_url):
+def test_plan(_url, _section):
```
jkonecny12, line 282: Why are these variables with `_` prefix?

test/wiki-report.py
```diff
@@ -0,0 +1,120 @@
...
+            if testcase["status"] == "pass":
+                passed_testcases.add(
+                    ResTuple(
+                        bot=True,
+                        cid=self.compose_id,
+                        compose=self.compose,
+                        dist="Fedora",
+                        env=testcase["env"],
+                        milestone="Rawhide",
+                        release="42",
```
jkonecny12, line 59: Please put the `TODO` from commit message to the code. It is more visible.

test/wiki-report.py
```diff
@@ -0,0 +1,120 @@
...
+        with open(self.report_path, "r") as f:
+            return json.load(f)
+
+    def login(self, wiki):
+        if not wiki.logged_in:
+            # This seems to occasionally throw bogus WrongPass errors
+            try:
+                wiki.login()
+            except mwclient.errors.LoginError:
+                wiki.login()
```
jkonecny12, line 81: Shouldn't we put at least a small sleep before the second attempt?

Also we should log why was the login unsuccessful if possible.

test/wiki-report.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.error("could not log in to wiki")
+            raise LoginError
+
+    def upload_report(self, wiki, passed_testcases):
+        # Submit the results
+        (insuffs, dupes) = wiki.report_validation_results(passed_testcases)
+
+        for dupe in dupes:
+            tmpl = "already reported result for test %s, env %s! Will not report dupe."
+            logger.info(tmpl, dupe.testcase, dupe.env)
```
jkonecny12, line 93: I don't think that `tmpl` variable is necessary, just write the message on two lines or create a new line before.

Just a suggestion, I don't insist on fixing this.

test/wiki-report.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.debug("full ResTuple: %s", dupe)
+
+        for insuff in insuffs:
+            tmpl = "insufficient data for test %s, env %s! Will not report."
```
jkonecny12, line 97: Same here.

test/wiki-report.py
```diff
@@ -0,0 +1,120 @@
...
+
+        passed_testcases = self.get_passed_testcases()
+        logger.info("passed testcases: %s", passed_testcases)
+
+        self.login(wiki)
+
+        self.upload_report(wiki, passed_testcases)
+
+
+
```
jkonecny12, line 114: Too many blank lines.

jkonecny12 (approved):
A few minor notes. Looks great to me overall.

test/wiki-report.py
KKoukiou: it would be long line

test/wiki-report.py
KKoukiou: Hm, not sure how it would help, we are guessing at this point. Anyway, I copied this code from https://pagure.io/fedora-qa/fedora_openqa/blob/main/f/src/fedora_openqa/report.py#_338

PR #602, opened 2025-01-21
Ensure that URL path is always set when loading the app

KKoukiou:
This fixes the 'Launch storage editor' button being disabled on Live ISO.

Resolves: rhbz#2336488


jkonecny12 (approved):
Looks good to me as much as I can tell.

PR #605, opened 2025-01-22
Parse hidden_webui_pages option from anaconda.conf

KKoukiou:
There was mismatch in the option name between front-end and back-end.


jkonecny12 (approved):
Looks good to me. Just fix the tests first please.

PR #607, opened 2025-01-23
tests: when testing against a compose checkout the TAG used in the compose

KKoukiou:
So that we are using for testing the same version of the Web UI source code (tests etc) that the anaconda-webui RPM has in the compose.


martinpitt (approved):
Thanks! Nice that this can be cleaned up now.

PR #609, opened 2025-01-23
tests: read expected hidden screens on Live ISO from anaconda.conf

KKoukiou:


M4rtinK (approved):
Looking good. :)

jkonecny12 (approved):

M4rtinK (approved):
Yep, that should be more future proof. :)

PR #611, opened 2025-01-23
tests: remove sleep before clicking 'next'

KKoukiou:
Next button is disabled when no disks are selected, we don't need the explicit sleep in the tests.


elkoniu (approved):
LGTM

PR #615, opened 2025-01-27
Fix tests after adding Welcome screen

jkonecny12:
Tests are now failing because pixel tests do not contain images with the Welcome screen.


adamkankovsky (approved):
That looks good, thanks for the change.

PR #631, opened 2025-02-11
tests: stop hardcoding the payload for fedora-rawhide only

KKoukiou:
Payload needs to be set per-image variant.


rvykydal (approved):
Looks good to me.

PR #634, opened 2025-02-12
tests: allow wiki-reporting even if test suite partially failed

KKoukiou:


M4rtinK (approved):
Looking good as far I can tell. :)

KKoukiou:
The failures are expected, compose runs are still not completely green.

PR #637, opened 2025-02-13
tests: fix old compose wiki reporting

KKoukiou:
When testing composes we checkout the git repository to the tag that anaconda-webui has in that compose. We do that so that the tests match the code tested.
However, there is some infra logic, that got more recently added, and is there for reporting to fedora wiki.

Let's copy the relevant files from master till these changes are part of the composes we are interested to test.

### DIFF
--- a/test/run
+++ b/test/run
@@ -42,13 +42,18 @@ if [ -n "${TEST_COMPOSE-}" ]; then
     COMPOSE_A_PACKAGES="https://kojipkgs.fedoraproject.org/compose/rawhide/$TEST_COMPOSE/compose/Everything/x86_64/os/Packages/a/"
     ANACONDA_WEBUI_TAG=$(curl -s $COMPOSE_A_PACKAGES | grep -oP '>anaconda-webui-[0-9]+' | cut -d "-" -f3)
 
-    # FIXME: Keep test/wiki-report.py from the current commit
-    # This is a workaround till test/wiki-report is part of the releases we are testing
-    cp test/wiki-report.py /tmp/wiki-report.py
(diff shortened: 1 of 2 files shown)


test/run
```diff
@@ -42,13 +42,18 @@ if [ -n "${TEST_COMPOSE-}" ]; then
...
-    # FIXME: Keep test/wiki-report.py from the current commit
-    # This is a workaround till test/wiki-report is part of the releases we are testing
-    cp test/wiki-report.py /tmp/wiki-report.py
+    # FIXME: Checkout only test/helpers directory and test/check-* files to the corresponding tag
+    # Other files are infra related and we need to keep them from main to utilize the wiki-report logic
+    git checkout $ANACONDA_WEBUI_TAG -- test/helpers test/check-* test/reference
 
-    git checkout $ANACONDA_WEBUI_TAG
+    # FIXME: Cherry-pick 325e09cba86bcb404dc62a7c82717c6e2f86f958 because test_plans decorator has changed in the meantime
+    git show  325e09cba86bcb404dc62a7c82717c6e2f86f958 -- test/check* | git apply -
```
elkoniu, line 50: Shall we set some git tag and reference it instead of a particular commit? Or this is also temporary?

elkoniu (commented):
LGTM, one minor comment.

test/run
elkoniu: Ok

elkoniu (approved):

rvykydal (approved):

KKoukiou:
The tokens are accessible only when the workflow is ran from a github action and not on a PR. This was expected to timeout.

PR #641, opened 2025-02-14
infra: wiki-report fixes

KKoukiou:


test/wiki-report.py
```diff
@@ -117,4 +117,11 @@ def run(self):
         print("Usage: compose-report.py <report_path>")
         sys.exit(1)
 
+    try:
+        with open("~/.openidc/oidc_wikitcms.json", "r") as f:
```
martinpitt, line 121: home dir is impractical for reasons -- please check `os.getenv("COCKPIT_FEDORA_WIKI_TOKEN")` and if it's set, then that is the path of the file.

martinpitt (changes requested):

test/wiki-report.py
martinpitt: Oh I see -- if it is set, please copy it to that destination path. We can't mount it as volume from outside with the correct permissions.

martinpitt: See https://github.com/cockpit-project/cockpituous/pull/642

test/wiki-report.py
```diff
@@ -117,4 +119,10 @@ def run(self):
         print("Usage: compose-report.py <report_path>")
         sys.exit(1)
 
+    if not os.getenv("COCKPIT_FEDORA_WIKI_TOKEN"):
+        print("Please set COCKPIT_FEDORA_WIKI_TOKEN environment variable with the file path to the wiki token.")
+        sys.exit(1)
+
+    shutil.copy(os.getenv("COCKPIT_FEDORA_WIKI_TOKEN"), "~/.openidc/oidc_wikitcms.json")
```
martinpitt, line 126: You most probably need an `os.makedirs()` here?

martinpitt: Also, I find it awkward to do this in `wiki-report.py` -- that's a file which you likely call as a developer locally, and then all of that is unexpected. Can you please move it to test/run?

martinpitt (changes requested):

martinpitt (approved):
LGTM!

PR #649, opened 2025-02-18
tests: move testmap for fedora wiki tests in a json file

KKoukiou:


KKoukiou (commented):
This makes it easier to isolate test code from test framework code, therefore easier to test older composes by keeping the test infra changes from master. (Example: mapping of test cases to new Fedora QA tests)

adamkankovsky (approved):
It looks fine to me. I think it's definitely more sustainable this way.

jkonecny12 (approved):
Looks good to me. I'm just not sure why we are removing those decorators. Seems to me like a better option to keep this information in the source code.

KKoukiou:
> Looks good to me. I'm just not sure why we are removing those decorators. Seems to me like a better option to keep this information in the source code.

The reason is that when we are testing against COMPOSES in order to report our test results to fedora wiki, we checkout an older version of the source code. However we would like to possibility to update the mapped test cases, for example if they add new test cases. 

https://github.com/rhinstaller/anaconda-webui/blob/main/test/run#L53

Having the test infra logic in the source code might be nice for readability but it's really bad if we want to update something for tests which are already in stable branches, (aka fedora-42)

KKoukiou:
Failing tests are unrelated, there is a bit of flakiness the last days,  i will deal with it right after freeze.

PR #650, opened 2025-02-18
tests: add support for testing Fedora 42 composes

KKoukiou:


jkonecny12:
We need to be able to minimize the hardcoded `42` everywhere but that is for future improvement.

adamkankovsky (approved):
Looks great to me.

PR #652, opened 2025-02-18
tests: fix: the first argument to next() must be an iterator

KKoukiou:
We were passing a list comprehension


adamkankovsky (approved):
I missed that in the original PR, thanks for the correction

PR #656, opened 2025-02-21
test: Dual booting with existing Fedora OS

adamkankovsky:
Use two different fedora installations to verify that the original EFI boot partition is preserved in the useFreeSpace scenario, so that both operating systems can boot.

### DIFF
--- a/test/helpers/operating_systems.py
+++ b/test/helpers/operating_systems.py
@@ -42,7 +42,7 @@ sector-size: 512
         self.machine.execute("echo '%s' | sfdisk /dev/vda" % self.WINDOWS_SFDISK)
 
 
-class DualBootHelper_E2E(VirtInstallMachineCase):
+class DualBootHelperDebian(VirtInstallMachineCase):
     def verifyDualBootDebian(self, root_one_size=None, root_two_size=None):
         b = self.browser
         m = self.machine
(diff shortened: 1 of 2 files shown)


KKoukiou:
Please use: `bots/tests-trigger 656  fedora-rawhide-boot/efi` && `bots/tests-trigger 656  fedora-42-boot/efi` for not running all unrelated tests during development in order to save some CI resources.

KKoukiou (commented):
Please do not create new file. Just put the new test in the existing test/check-storage-use-free-space-e2e

Storage files are split - per - scenario.

also run_boot decorator can just specify run_boot("efi") as this issue exists only in EFI boots

test/check-storage-reuse-efi-e2e
```diff
@@ -0,0 +1,58 @@
...
+
+
+class TestUseFreeSpace_FedoraDualBoot_E2E(DualBootHelperFedora_E2E, VirtInstallMachineCase):
+    """
+    Installs Fedora alongside an existing Fedora using 'Use free space' and checks both boot.
+    """
+
+    @disk_images([("fedora-rawhide", 20)])
+    @run_boot("bios", "efi")
+    def test_use_free_space_dual_fedora(self):
```
KKoukiou, line 33: Naming conventions: method names use camel case.

KKoukiou (changes requested):

test/helpers/operating_systems.py
```diff
@@ -89,3 +89,65 @@ def verifyDualBootDebian(self, root_one_size=None, root_two_size=None):
             self.assertEqual(vda15["mountpoints"], ["/boot/efi"])
         else:
             self.assertEqual(vda15["mountpoints"], ["/efi", "/boot/efi"])
+
+
+class DualBootHelperFedora_E2E(VirtInstallMachineCase):
```
KKoukiou, line 94: E2E can be removed. It's just naming convention for test classes.

Also you can adjust a few lines above DualBootHelper to be DualBootHelperDebian, and make sure there is not some methods you could pottentially reuse.

jkonecny12:
Closing with explanation in https://bugzilla.redhat.com/show_bug.cgi?id=2324913

PR #658, opened 2025-02-24
 storage: cockpit-storage: select newly created mdarrays if the user created them

KKoukiou:


rvykydal (approved):
Looks good to me.

src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -430,6 +432,66 @@ const CheckStorageDialog = ({
...
+            }
+            return true;
+        });
+        // TODO: Consider moving this logic to the backend
+        if (mdArraysNotSupported.length > 0) {
+            setError({
+                message: cockpit.format(
+                    _(
+                        "Invalid RAID configuration detected.\n" +
+                        "If your RAID array is created directly on top of disks, a partition table must be created on the array.\n" +
```
KKoukiou, line 478: @vojtechtrefny we discussed this last week, I think we should move this logic elsewere, (blivet and fedora documentation)
But for resolving fast the bug, can you please review this error message?

KKoukiou:
@rvykydal added one more test with the corresponding code logic for detecting invalid raid configuration.

src/components/storage/CockpitStorageIntegration.jsx
vojtechtrefny: Strictly speaking, the first part is not true -- we support disks with filesystem without partition table, it's just not something generally recommend (and doesn't make much sense during installation anyway). But I am not really sure what will Anaconda do with disks like that (I think default behaviour is to create a new partition table on all selected disks) so I guess it makes sense to mark it as unsupported.

vojtechtrefny (approved):
Looks good to me.

test/helpers/storage.py
```diff
@@ -82,8 +82,10 @@ def check_constraint(self, constraint, required=True):
     def confirm_entering_cockpit_storage(self):
         self.browser.click("#cockpit-storage-integration-enter-storage-confirm")
 
-    def return_to_installation(self):
+    def return_to_installation(self, error=None):
```
rvykydal, line 85: I am surprised I don't see the error set anywhere in the added tests.

KKoukiou: Omg I forgot. Repushing.

KKoukiou:
The remaining test failures are unrelated.

PR #661, opened 2025-02-26
tests: add two new test scenarios for splitting test runs even more

KKoukiou:


martinpitt (approved):
Thanks! LGTM, but please make sure this doesn't cause new test failures (hard for me to tell which ones are "expected").

test/run
```diff
@@ -35,8 +35,14 @@ case "${TEST_SCENARIO:=}" in
     *expensive*)
         RUN_OPTS="$(echo "$ALL_TESTS" | grep -E "$RE_EXPENSIVE")"
         ;;
+    *cockpit*)
```
mvollmer, line 41: This way of splitting the tests of course has the risk of putting some tests into more than one scenario. For example, there might very well be a test names TestStorageE2E which would end up both in the "expensive" and "storage" scenario. We have the same issue in Cockpit's own test/run and it would probably be good to find a better approach.

But I guess this works for now.

test/check-storage-use-free-space-e2e
```diff
@@ -23,7 +23,7 @@ from storage import Storage
 from testlib import test_main  # pylint: disable=import-error
 
 
-class TestUseFreeSpace_E2E(DualBootHelper_E2E, VirtInstallMachineCase):
+class TestStorageUseFreeSpace_E2E(DualBootHelper_E2E, VirtInstallMachineCase):
```
mvollmer, line 26: Ah! Doesn't this end up in _both_ "expensive" and "storage"?

test/run
KKoukiou: OH true, good catch. we do have such tests. I will adjust to exclude the expensive tests in the storage and cockpit scenarios.

mvollmer (approved):

PR #663, opened 2025-02-27
localization: spawn gnome settings for changing keyboards in workstation

KKoukiou:


anaconda-gnome-control-center.desktop
```diff
@@ -0,0 +1,16 @@
+[Desktop Entry]
```
jkonecny12, line 1: I wonder if Workstation SIG will be happy about new application in the Gnome Shell like this?
Please take into account that this application will be available also in the installed system.

src/components/localization/Keyboard.jsx
```diff
@@ -64,3 +64,37 @@ export const KeyboardSelector = ({ idPrefix }) => {
         </FormSelect>
     );
 };
+
+export const KeyboardGnome = () => {
+    const [compositorSelectedLayout, setCompositorSelectedLayout] = useState(null);
+
+    useEffect(() => {
+        const onFocus = () => getCompositorSelectedLayout().then(setCompositorSelectedLayout);
```
jkonecny12: This will not work. Why do we have it here?

src/components/localization/Keyboard.jsx
```diff
@@ -64,3 +64,37 @@ export const KeyboardSelector = ({ idPrefix }) => {
...
+
+        onFocus();
+
+        window.addEventListener("focus", onFocus);
+        return () => window.removeEventListener("focus", onFocus);
+    }, []);
+
+    return (
+        <Flex alignItems="center" flexWrap={{ default: "nowrap" }}>
+            <Text>{compositorSelectedLayout || _("Unknown layout")}</Text>
```
jkonecny12: Another API which doesn't work.

jkonecny12 (commented):
Could you please first answer my question below.

anaconda-gnome-control-center.desktop
KKoukiou: Yes this is a problem, but I dont think there is another way. 
cockpit.spawn did not work, as it's not setting the env variables properly. I can retry implementing this with cockpit-spawn.

src/components/localization/Keyboard.jsx
KKoukiou: I think you re-viewed an outdated version?

anaconda-gnome-control-center.desktop
jkonecny12: We should ping Workstation SIG about this. I'll reach them out.

src/apis/localization.js
```diff
@@ -151,14 +151,61 @@ export const getCompositorSelectedLayout = () => {
...
+    return setProperty("XLayouts", cockpit.variant("as", layouts));
+};
+
+export const populateMissingKeyboardConfiguration = async ({ onFail, onSuccess }) => {
+    const task = await callClient("PopulateMissingKeyboardConfigurationWithTask");
+
+    const taskProxy = new LocalizationClient().client.proxy(
+        "org.fedoraproject.Anaconda.Task",
+        task
+    );
```
jkonecny12, line 172: You should be fine to remove this API. It's not used anywhere.

jkonecny12 (approved):
Looks good to me otherwise.

KKoukiou:
Unrelated test failures.

PR #668, opened 2025-03-04
tests: Switch from cp to rsync for system copy

KKoukiou:
Use rsync instead of cp to exclude unnecessary paths (/dev, /proc, /sys, etc.), ensuring a more reliable system transfer, as cp got stuck often.


KKoukiou:
This indeed fixes the flaky home reuse encrypted test: will merge as part of https://github.com/rhinstaller/anaconda-webui/pull/658

PR #669, opened 2025-03-04
tests: adjust tests because of new packages in rawhide ISO 

KKoukiou:


jkonecny12 (approved):
Looks good to me AFAIK

PR #683, opened 2025-03-10
tests: remove unnecessary class wrappers

KKoukiou:
Since bd2b83a730c97e4a72ed62475b0b49e7dd955301, that we introduced the disk_images test method decorator instead of the class attribute for specifing disks we can just have one wrapper class per file, making the test files slightly cleaner.

Pixel test references had to be updated with the new names.


KKoukiou:
@rvykydal what do you think about this? I prefer I think single class per-file.

rvykydal (approved):
Looks good to me.

PR #688, opened 2025-03-11
Enhance Frontend Exception Reporting

KKoukiou:
Frontend tracebacks shown in Bugzilla lacked source maps, as window.onerror does not display errors with applied source maps. To address this, integrate the stacktrace-js library to improve the readability of reported frontend exceptions.


KKoukiou:
@bruno-fs Obviously after holiday - I would appreciate a second look.

PR #689, opened 2025-03-11
tests: pre-create images pool otherwise parallel VM spawn might race on creating it

KKoukiou:
This avoids the following error:
ERROR    Validating install media '...' failed: Could not define storage pool: operation failed: \n
pool 'images' already exists with uuid ...

### DIFF
--- a/test/run
+++ b/test/run
@@ -92,6 +92,9 @@ if [ -z "${TEST_OS-}" ]; then
   TEST_OS=fedora-rawhide-boot
 fi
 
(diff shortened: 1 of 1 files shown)


KKoukiou:
Ehm now it's trying to create images-1

ERROR    Validating install media '/work/make-checkout-workdir/bots/images/fedora-rawhide-boot' failed: Could not define storage pool: operation failed: pool 'images-1' already exists with uuid 6365e93d-ea47-47a7-88bc-4a56c73d09da

PR #695, opened 2025-03-13
tests: add support for reporting tests on either staging or production wiki

KKoukiou:
Introduce this knowledge in the "compose" scenario. It can now be: TEST_SCENARIO=compose-<compose_id> or
TEST_SCENARIO=compose-<compose_id>-staging

Also make test/wiki-report script configurable on which wiki instance to report to.

Lastly introduce a parameter in the Workflow to be able to choose the wiki instance there as well.


test/run
```diff
@@ -111,9 +117,14 @@ if [ -n "${TEST_COMPOSE-}" ] && [ "$TESTING_BRANCH" == "main" ]; then
     fi
 
     mkdir -p ~/.openidc
-    cp $COCKPIT_FEDORA_WIKI_STAGING_TOKEN ~/.openidc/oidc_wikitcms.json
+    if [ -n "${TEST_FEDORA_WIKI_STAGING-}" ]; then
```
martinpitt, line 120: Please double-check this. This check means "variable is defined", but that's not what you do in line 39/42. Are you sure you don't mean `if [ "${TEST_FEDORA_WIKI_STAGING-}" = "true" ]`?

test/run
```diff
@@ -111,9 +117,14 @@ if [ -n "${TEST_COMPOSE-}" ] && [ "$TESTING_BRANCH" == "main" ]; then
...
+        cp $COCKPIT_FEDORA_WIKI_STAGING_TOKEN ~/.openidc/oidc_wikitcms.json
+        test/wiki-report -t test/report.json
```
martinpitt, line 122: I suggest to spell this out as `--testing` (or better `--staging`, see below).

test/wiki-report.py
```diff
@@ -130,8 +131,9 @@ def run(self):
 
 
 if __name__ == "__main__":
-    if len(sys.argv) != 2:
-        print("Usage: compose-report.py <report_path>")
-        sys.exit(1)
+    parser = argparse.ArgumentParser(description="Report test results to the Fedora wiki")
+    parser.add_argument("report", help="Path to the JSON report file")
+    parser.add_argument("--testing", "-t", action="store_true", help="Operate on the staging wiki (for testing)")
```
martinpitt, line 136: How about `--staging`?

test/run
KKoukiou: Did the change to:
 if [ "$TEST_FEDORA_WIKI_STAGING" == "true" ]; then

It's always defined for COMPOSE scenario.

test/run
```diff
@@ -111,9 +117,14 @@ if [ -n "${TEST_COMPOSE-}" ] && [ "$TESTING_BRANCH" == "main" ]; then
     fi
 
     mkdir -p ~/.openidc
-    cp $COCKPIT_FEDORA_WIKI_STAGING_TOKEN ~/.openidc/oidc_wikitcms.json
+    if [ "$TEST_FEDORA_WIKI_STAGING" == "true" ]; then
```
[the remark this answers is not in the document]
KKoukiou: Aha I will never know what's bash and what's plain

test/run
KKoukiou: Your suggestion was put in place in the end :)

PR #698, opened 2025-03-13
tests: Map some non-E2E test cases to the wiki test map

KKoukiou:
Non-E2E tests validate all expected checks without rebooting into the installed system, avoiding resource overhead. Instead, they verify the `fstab` file and the UI indicating partition changes (removal/resizing), providing assurance that the created storage layout aligns with expectations.

### DIFF
--- a/test/wiki-testmap.json
+++ b/test/wiki-testmap.json
@@ -22,6 +22,10 @@
                 {
                     "testname": "TestStorageHomeReuse_E2E.testBasic",
                     "fedora-wiki-testcase": "QA:Testcase_partitioning_webui_guided_btrfs_preserve_home"
+                },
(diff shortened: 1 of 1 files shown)


KKoukiou:
@adamwill Hey Adam,
We've mapped some non-E2E test cases to the wiki test map. These tests perform all expected checks but skip rebooting into the installed system to save resources.

Does this approach align with Fedora QA's testing strategy? Let me know your thoughts. Thanks!

AdamWill:
I don't *love* it, but tbf not all the openQA storage tests actually check that the end result is correct, so they're not perfect either. So we can probably live with it.

jkonecny12 (approved):

PR #703, opened 2025-03-17
Fix translations that were prepared as string concatenation

KKoukiou:
This was generating some funny and broken translations.

Relevant to: rhbz#2351026

### DIFF
--- a/src/components/storage/CockpitStorageIntegration.jsx
+++ b/src/components/storage/CockpitStorageIntegration.jsx
@@ -149,9 +149,7 @@ const CockpitStorageConfirmationModal = ({ handleCancelOpenModal, handleConfirmO
         >
             <TextContent>
                 <Text>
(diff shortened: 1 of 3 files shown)


jkonecny12:
I agree with the comment on BZ. Such a long line is pretty awful. Couldn't we have middle-ground for this?

KKoukiou:
> I agree with the comment on BZ. Such a long line is pretty awful. Couldn't we have middle-ground for this?

I dont think there is a way in cockpit.gettext to currently parse any kinda of string concatenation.

martinpitt (approved):
Thanks. Indeed `xgettext` cannot parse concatenated strings or anything similarly funny. If you need that, that'd be some work in the translation extraction. But until that exists, this is correct and fixes i18n. Thanks!

jkonecny12 (approved):
That is lame but I guess we don't have an option.

PR #706, opened 2025-03-18
storage: cockpit-storage-integration: do not allow proceeding with invalid configuration / allow bootloader on seperate disk

KKoukiou:


src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -509,7 +510,7 @@ const CheckStorageDialog = ({
     useEffect(() => {
         // If the required devices needed for manual partitioning are set up,
         // and prepare the partitioning
-        if (checkStep !== "prepare-partitioning") {
+        if (checkStep !== "prepare-partitioning" || appliedPartitioning) {
```
rvykydal, line 508: Thinking out loud: do I understand it right that now that we reset partitioning on each rescan we need this added guard (non existing appliedPartitioning) to prevent preparing (applying) partitioning multiple times?

rvykydal: Oh wait, I was looking at the patch in isolation, I guess I should look at the while first.

src/apis/storage_partitioning.js
```diff
@@ -248,14 +254,31 @@ export const gatherRequests = ({ partitioning }) => {
...
         await partitioningSetEncrypt({ encrypt: luks.encrypted, partitioning });
     }
     if (luks?.passphrase) {
         await partitioningSetPassphrase({ partitioning, passphrase: luks.passphrase });
     }
 
+    const method = await getPartitioningMethod({ partitioning });
+    if (method === "MANUAL") {
+        const requests = await gatherRequests({ partitioning });
+        const bootloaderDevice = requests.find(request => ["biosboot", "efi"].includes(request["format-type"].v))?.["device-spec"].v;
```
rvykydal, line 268: I wonder if we should use [bootloaderTypes](https://github.com/rhinstaller/anaconda-webui/blob/06271e44e0fda0cfdaa2c1c6be36f75eff259611/src/helpers/storage.js#L176) here but don't insist - I guess it is not the only place we work just with these two types.

src/helpers/storage.js
```diff
@@ -139,6 +139,18 @@ export const getParentPartitions = (deviceData, device) => {
     return deviceData[device].parents.v.map(parent => getParentPartitions(deviceData, parent)).flat(Infinity);
 };
 
+/* Get the parent disk ID of a given device
+ * @param {Object} deviceData - The device data object
+ * @param {string} device - The ID of the device
+ * @returns {string}
+ * */
+export const getDeviceDisk = (deviceData, device) => {
```
rvykydal, line 147: [Feel free to ignore] I wonder if we should use the helper here: https://github.com/rhinstaller/anaconda-webui/blob/06271e44e0fda0cfdaa2c1c6be36f75eff259611/src/components/storage/MountPointMapping.jsx#L274. If so I guess it is out of scope of this PR anyway.

src/hooks/Storage.jsx
```diff
@@ -148,7 +145,6 @@ export const getNewPartitioning = async ({
     if (method === "AUTOMATIC") {
         await setInitializeLabelsEnabled({ enabled: true });
     }
-    await setBootloaderDrive({ drive: "" });
```
rvykydal, line 151: I'll try to think a bit if not setting it here can't break creating the automatic partitioning.

test/check-storage-cockpit-e2e
```diff
@@ -115,6 +115,85 @@ class TestStorageCockpitIntegration_E2E(VirtInstallMachineCase, StorageCase):
...
+        self.dialog({"type": "gpt"})
+        self.click_dropdown(self.card_row("Storage", 2), "Create partition")
+        if self.is_efi:
+            self.dialog({"size": 100, "type": "efi", "mount_point": "/boot/efi"})
+        else:
+            self.dialog({"size": 1, "type": "biosboot"})
+        self.click_dropdown(self.card_row("Storage", 3), "Create partition")
+        self.dialog({"size": 1070, "type": "ext4", "mount_point": "/boot"})
+
+        # Create biosboot, /boot and / partitions on the RAID device
```
rvykydal, line 155: The comment is out of sync?

rvykydal (approved):
Looks good to me.
None of my comments is blocking.

src/components/storage/CockpitStorageIntegration.jsx
KKoukiou: Yes, if there is applied partitioning, we are sure that we are past this step.

src/apis/storage_partitioning.js
KKoukiou: Good point. I wiill.

src/hooks/Storage.jsx
KKoukiou: It's now done as part of the applyStorage (src/apis/storage_partitioning.js), so for automatic the behavior is the same.

test/check-storage-cockpit-e2e
KKoukiou: Will fix.

src/helpers/storage.js
KKoukiou: Let's fix that for rawhide

KKoukiou:
Windows test unrelated failure.

src/hooks/Storage.jsx
rvykydal: I put it wrong, I was thinking if creating the partitioning (CreatePartitioning) does not require it - vs configuring the partitioning (ConfigurePartitioning).
(Actually the concern is not regarding only automatic it was misleading in my question.) But given the value is "" I guess not.

PR #707, opened 2025-03-18
storage: cockpit-storage: prevent bootloader partitions on MDRAID devices

KKoukiou:
Adjust existing test to cover for both biosboot end uefi boot modes.

Resolves: rhbz#2352953

### DIFF
--- a/src/components/storage/CockpitStorageIntegration.jsx
+++ b/src/components/storage/CockpitStorageIntegration.jsx
@@ -66,6 +66,7 @@ import { getDevicesAction, setStorageScenarioAction } from "../../actions/storag
 
 import { debug } from "../../helpers/log.js";
 import {
+    bootloaderTypes,
     getDeviceAncestors,
     getDeviceByName,
     getDeviceByPath,
@@ -305,7 +306,7 @@ export const preparePartitioning = async ({ devices, newMountPoints, selectedDis
         });
 
         await setManualPartitioningRequests({ partitioning, requests });
-        return partitioning;
+        return [partitioning, requests];
     } catch (error) {
         console.error("Failed to prepare partitioning", error);
     }
@@ -540,7 +541,24 @@ const CheckStorageDialog = ({
             // CLEAR_PARTITIONS_NONE = 0
             try {
                 await setInitializationMode({ mode: 0 });
-                const partitioning = await preparePartitioning({ devices, newMountPoints, selectedDisks });
+                const [partitioning, requests] = await preparePartitioning({ devices, newMountPoints, selectedDisks });
+
+                // FIXME: Do not allow stage1 device to be mdarray when this was created in Cockpit Storage
+                // Cockpit Storage creates MDRAID with metadata 1.2, which is not supported by bootloaders
+                // See more: https://bugzilla.redhat.com/show_bug.cgi?id=2355346
+                const bootloaderRequest = requests.find(request => bootloaderTypes.includes(request["format-type"].v));
+                // PMBR does not have a bootloader necessarily
+                const bootloaderDevice = bootloaderRequest?.["device-spec"].v;
+                const bootloaderDriveMDRAID = bootloaderDevice && getDeviceAncestors(devices, bootloaderDevice).find(device => devices[device].type.v === "mdarray");
+                if (bootloaderDriveMDRAID) {
+                    throw Error(
+                        cockpit.format(
+                            _("'$0' partition on MDRAID device $1 found. Bootloader partitions on MDRAID devices are not supported."),
+                            bootloaderRequest["format-type"].v,
+                            devices[bootloaderDriveMDRAID].name.v
+                        )
+                    );
+                }
 
                 applyStorage({
                     devices,
--- a/test/check-storage-cockpit
+++ b/test/check-storage-cockpit
@@ -491,10 +491,11 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     @disk_images([("", 15), ("", 15), ("", 15)])
-    def testRAIDUnsupportedMetadataCheck(self):
+    @run_boot("bios", "efi")
+    def testUnsupportedBootloaderOnMDRAID(self):
         b = self.browser
         m = self.machine
-        i = Installer(b, m)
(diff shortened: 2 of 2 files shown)


KKoukiou:
> Boot partitions on RAID are supported, it just needs to use either 0.90 or 1.0 version of MD metadata instead of the default 1.2. Blivet make sure 1.0 metadata version is used when creating RAID for boot mount points.

Cockpit-storage can't create MDRAID with the supported metadata yet https://bugzilla.redhat.com/show_bug.cgi?id=2353304 

Let's prevent using bootloaders on MDRAID till this is possible

rvykydal (approved):
Looks good to me.

src/components/storage/CockpitStorageIntegration.jsx
```diff
@@ -488,6 +488,20 @@ const CheckStorageDialog = ({
...
 
+        // FIXME: Do not allow stage1 device to be mdarray when this was created in Cockpit Storage
+        // Cockpit Storage creates MDRAID with metadata 1.2, which is not supported by bootloaders
+        // See more: https://bugzilla.redhat.com/show_bug.cgi?id=2353304
+        const bootloaderDevices = Object.keys(devices).filter(device => bootloaderTypes.includes(devices[device].formatData.type.v));
+        if (
+            bootloaderDevices.some(bootloaderDevice => getDeviceAncestors(devices, bootloaderDevice).some(ancestor => mdArrays.includes(ancestor)))
+        ) {
+            setError({
+                message: _("Bootloaders are not supported on MDRAID devices. To ensure the system can boot, avoid placing the bootloader on MDRAID.")
```
garrett, line 499: We need to change this message a bit, and I think we might need to tailor it for EFI vs. biosboot as well, as — at least as I understand it — the problem isn't just with /boot, but also with /boot/efi and biosboot. Since we need to tell them what to do to resolve this issue, it needs to be specific to each type.

(You wouldn't want to tell someone to have to do something about the EFI partition on a BIOS system, and vice versa.)

garrett (dismissed):

src/components/storage/CockpitStorageIntegration.jsx
garrett: (I need a little more clarification on what exactly needs to be done before suggesting a revised message. But I'm happy to work on suggesting a message ASAP.)

rvykydal: Thank you for reviewing the message, I wasn't sure about it  wanted to suggest going with just the first sentence, something like 'placing boot loader on MDRAID device is not supported' but it would be perhaps too vague.

mvollmer:
What about disabling MDRAID creation in Cockpit instead until it can create MDRAIDs with the right metadata version? Then we don't need to spend effort on the messaging here.

(https://github.com/cockpit-project/cockpit/pull/21730)

garrett:
> What about disabling MDRAID creation in Cockpit instead until it can create MDRAIDs with the right metadata version? Then we don't need to spend effort on the messaging here.

I'm fine with either approach, as long as we're doing what needs to be done in Anaconda. Whatever approach you two think is best, I'll do what I can to assist.

mvollmer:
And is if we wouldn't have enough decisions to make already, I have also prepared https://github.com/cockpit-project/cockpit/pull/21731. That PR will make Cockpit create mdraids with metadata version 1.0 when in Anaconda mode.

garrett:
> I have also prepared https://github.com/cockpit-project/cockpit/pull/21731. That PR will make Cockpit create mdraids with metadata version 1.0 when in Anaconda mode.

@mvollmer: Oh! I was actually wondering if that was possible and how much effort that'd take. But I assumed that it wouldn't be possible or that it'd be difficult...

If I understand it correctly, that's my preference. No UI adapt, no error; everything should "just work". It should solve the issue in a clean way. I'd think that PR would be better than the other options.

🎉🎉🎉 Excellent! Thank you so much for working on this! 🎉🎉🎉

mvollmer:
> > I have also prepared [cockpit-project/cockpit#21731](https://github.com/cockpit-project/cockpit/pull/21731). That PR will make Cockpit create mdraids with metadata version 1.0 when in Anaconda mode.
> 
> @mvollmer: Oh! I was actually wondering if that was possible and how much effort that'd take. But I assumed that it wouldn't be possible or that it'd be difficult...

My my concern with this is the urgency, uncertain coverage of our automated tests, and limited time for manual testing. It's a completely new and different code-path for creating mdraids when in Anaconda mode, and subtle bugs might lurk.
We would have to run all our normal mdraid tests also in Anaconda mode now, or use the madadm code path always, not just when UDisks2 is insufficient...

KKoukiou:
With https://github.com/cockpit-project/cockpit/pull/21731 and https://github.com/rhinstaller/anaconda/pull/6290 we don't need this limitation here.

KKoukiou:
Let's have this as temporary fix, because the Anaconda backend error, will talk about invalid metadata version (1.2) incompatible with bootloaders. Since cockpit-storage does not expose any option to set RAID metadata, it would be confusing for users to get an error about something they can't change. 

Catch invalid setups earlier with this patch.

rvykydal (approved):
Looks good to me, but I think @garrett's comment about the message for user still holds.

To provide something more useful fast (without @garrett's help which would be better) maybe we should just use the type of the partition we found (it is one of `["efi", "biosboot", "appleboot", "prepboot"]`) in the message. Something like
`{type} partition on MDRAID device {device.name} found. Bootloader partitions on MDRAID devices are not supported.`.
OTOH generic message can be safer at this point if more hints are can be available elsewhere.

KKoukiou:
> Looks good to me, but I think @garrett's comment about the message for user still holds.
> 
> To provide something more useful fast (without @garrett's help which would be better) maybe we should just use the type of the partition we found (it is one of `["efi", "biosboot", "appleboot", "prepboot"]`) in the message. Something like `{type} partition on MDRAID device {device.name} found. Bootloader partitions on MDRAID devices are not supported.`. OTOH generic message can be safer at this point if more hints are can be available elsewhere.

OK I will send a followup. Also the existing commit I realized checks the while device tree, not only the request object. Adjusting to reflect that as well.

rvykydal (approved):
Looks good to me, thank you.

PR #712, opened 2025-03-20
Migrate isFinal to new ProductData API

adamkankovsky:
WebUI folowup for new ProductData API:[6284]( https://github.com/rhinstaller/anaconda/pull/6284)

### DIFF
--- a/src/apis/runtime.js
+++ b/src/apis/runtime.js
@@ -82,11 +82,26 @@ export class RuntimeClient {
 }
 
 /**
- *
+ * @returns {Promise}           Returns the full product data
+ */
+const getProductData = () => {
+    return getProperty("ProductData");
+};
+
+/**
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
This will break fedora-42 tests as the API does not exist for Fedora 42 but we still test that.

I would say to introduce a temporarary midd;e state with previous and new API used.

```

export const getIsFinal = async () => {
    try {
          return getProperty("IsFinal");
    } catch (ex) {
    if (e.name === "org.freedesktop.DBus.Error.UnknownMethod") {
          const productData = await getProductData();
          return productData.is_final_release;
     }
};
 
```

Check what kind of exception is thrown might now be that exactly.

KKoukiou (approved):
Did a clean rebase.

PR #714, opened 2025-03-20
storage: support disks with filesystem directly on them for installation

KKoukiou:
Relevant to: rhbz#2353524

### DIFF
--- a/src/components/review/StorageReview.jsx
+++ b/src/components/review/StorageReview.jsx
@@ -169,14 +169,14 @@ const DeviceRow = ({ disk }) => {
     const newMountPointRows = Object.entries(mountPoints).filter(mp => {
         const parents = getDeviceAncestors(devices, mp[1]);
 
-        return parents.includes(disk);
+        return parents.includes(disk) || mp[1] === disk;
     });
 
     const swap = Object.keys(devices).find(device => devices[device].formatData.type.v === "swap");
     if (swap) {
         const parents = getDeviceAncestors(devices, swap);
 
-        if (parents.includes(disk)) {
+        if (parents.includes(disk) || swap === disk) {
             newMountPointRows.push(["swap", swap]);
         }
     }
@@ -193,8 +193,9 @@ const DeviceRow = ({ disk }) => {
 
         const parents = getDeviceAncestors(originalDevices, action["device-id"].v);
 
-        return parents.includes(disk);
+        return parents.includes(disk) || action["device-id"].v === disk;
     });
+
     return (
         <div>
             <span id={`disk-${disk}`}>{cockpit.format_bytes(deviceData.size.v)} {disk} {"(" + deviceData.description.v + ")"}</span>
--- a/src/helpers/storage.js
+++ b/src/helpers/storage.js
@@ -212,8 +212,9 @@ export const getUsableDevicesManualPartitioning = ({ devices, selectedDisks }) =
             return false;
         }
 
-        // All device's disks have to be in selected disks.
-        return selectedDisks.some(disk => ancestors.includes(disk));
+        // All device's parent disks have to be in selected disks.
+        // Or the device itself has to be in selected disks.
+        return selectedDisks.some(disk => [device, ...ancestors].includes(disk));
     });
 
     return usableDevices;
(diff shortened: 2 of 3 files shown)


rvykydal:
Maybe the layout of the review info could be improved. On the first sight I'm thinking it would be nice to have the mount info aligned somehow, but I don't insist. It might not be so easy and it can be quite an edge case to bother (at this point).
Screenshot from 2025-03-24 09-03-48

rvykydal:
> As blivet GUI did not allow this, this was never tested / exposed. Not sure what happens if the user pre-prepares a disk with this layout, before booting anaconda.
> 

I wonder if it is possible to achieve this by kickstart.

Yes, I think we should at least check that installer does not have serious issues with such a layout.

rvykydal:
> > I have some concerns about supportability of the layout asking just to be sure, so we don't have too much maintenance burden with systems with such layout produced - from installed system, but mainly for example for further installations on systems with such layout - be it erase all or reinstallation..
> 
> If we don't want to support such layouts, we should adjust the Validation checks in the storage to not allow this. We should not introduce permanent logic in the front end to dissallow specific layouts.
> 

Ah, I was thinking it was the backend check that prevents the layout in the rhbz#2353524.

> @vojtechtrefny says it's valid, it's just that blivet GUI did not expose formatting disks directly.
> 
> So, I am all in restricting what we support, but that needs to be done directly in backend and affected non UI installations as well.

If it is already possible by non-ui installations I am ok with not restricting it in UI (in the cockpit storage use case).

rvykydal:
> > > I have some concerns about supportability of the layout asking just to be sure, so we don't have too much maintenance burden with systems with such layout produced - from installed system, but mainly for example for further installations on systems with such layout - be it erase all or reinstallation..
> > 
> > 
> > If we don't want to support such layouts, we should adjust the Validation checks in the storage to not allow this. We should not introduce permanent logic in the front end to dissallow specific layouts.
> 
> Ah, I was thinking it was the backend check that prevents the layout in the rhbz#2353524.

Obviously I was wrong, looking at the patch.

vojtechtrefny:
> @vojtechtrefny says it's valid, it's just that blivet GUI did not expose formatting disks directly.

It's valid, we just don't recommend that because it makes future changes harder (it's already complicated to for example shrink existing storage to make some more space for `/home` or second installation with partitions or LVM and formatting the entire disk to a single filesystem makes it impossible, but there are people that want to do that).

> As blivet GUI did not allow this, this was never tested / exposed. Not sure what happens if the user pre-prepares a disk with this layout, before booting anaconda.

I think we will just remove the format and create partition table on the disk. We do that with all "empty" disks that are selected on the storage spoke and in this case we probably consider the disk to be empty.

KKoukiou:
> > @vojtechtrefny says it's valid, it's just that blivet GUI did not expose formatting disks directly.
> 
> It's valid, we just don't recommend that because it makes future changes harder (it's already complicated to for example shrink existing storage to make some more space for `/home` or second installation with partitions or LVM and formatting the entire disk to a single filesystem makes it impossible, but there are people that want to do that).
> 
> > As blivet GUI did not allow this, this was never tested / exposed. Not sure what happens if the user pre-prepares a disk with this layout, before booting anaconda.
> 
> I think we will just remove the format and create partition table on the disk. We do that with all "empty" disks that are selected on the storage spoke and in this case we probably consider the disk to be empty.

Since cockpit-storage is a generic editor and not bound to anaconda, we will not be expecting cockpit-storage to introduce extra logic for anaconda. If we dont want to support a partitioning layout, we should handle that in backend to give an error. 
What we also can do it add 'warnings' for suboptimal layouts like this one, in the backend, show it in frontend as a hint, and allow the user to react or not.
All these for f43. For Fedora-42 we need to unbreak this I think, by just making sure it either works or it's blocked (which we can only do through cockpit-storage)
I prefer the former, so this PR, as I dont want to introduce more hacks to cockpit storage.

KKoukiou:
Did a plain rebase so see latest results again.

vojtechtrefny:
> What we also can do it add 'warnings' for suboptimal layouts like this one, in the backend, show it in frontend as a hint, and allow the user to react or not.

We already show warnings for things like missing swap (or at least I remember seeing that on RHEL, it's probably not a warning on Fedora with zram) so the logic is already in the backend somewhere.

> All these for f43.

Yes, of course.

rvykydal (approved):
Looks good to me. Thank you for the update.

PR #718, opened 2025-03-21
[DBG IGNORE] Start debugging test

rvykydal:


KKoukiou:
@rvykydal please trigger with: bots/tests-trigger 718 fedora-rawhide-boot/storage

To avoid triggering the while suite and eating CI resources. Put [no-test] label for that.

PR #728, opened 2025-03-25
storage: fix parsing of usable devices for manual partitioning

KKoukiou:
Extended partitions appear to not have any children but are not leafs.

Added also a test for this.


test/check-storage-reclaim
```diff
@@ -236,7 +236,7 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
                 ("", "extended"),
                 ("2.5GB", "logical"),
             ],
-            is_efi=True
+            is_mbr=True
```
rvykydal, line 239: Shouldn't this be `is_mbr=False` ?

KKoukiou: extended partitions are only valid as part of MBR, this code path in partition_disk, was setting up msdos on disk instead of GPT and then supported creating extended partitions on this.

Btw I did not change the logic, I just changed the variable name, as it was missleading.

rvykydal (commented):
Apart from the question looks good to me.

KKoukiou:
> Apart from the question looks good to me.

See https://github.com/rhinstaller/anaconda-webui/pull/728#discussion_r2013735318

test/check-storage-reclaim
rvykydal: Okay, I was probably confused by the test name containing EFI compared to the newly added `_testExtendedPartition_partition_disk`

PR #729, opened 2025-03-25
tests: add check for /boot on RAID0 (FAILING rhbz#2354805)

KKoukiou:
https://bugzilla.redhat.com/show_bug.cgi?id=2354805


KKoukiou:
This should not be supported: https://github.com/rhinstaller/anaconda/pull/6311

PR #736, opened 2025-03-27
Migration to Patternfly v6

adamkankovsky:


KKoukiou (commented):
Great progress. I see the following todo items:
* Talk with @garrett about the overrides for Fedora styling
* Fix the test selectors so that tests pass
* Check all new pixel test screenshots and make sure that all changes are expected (we can all help with this)

garrett:
I've pushed up a couple of small fixes. I spent a lot of today trying to:

1. Figure out why there's a grey strip on the bottom of the page. I tried all sorts of stuff, but I can't see why that's there. Are things nested properly? I can't find anything that seems to make that happen... it's so odd. I eventually gave up (for now) and moved on to branding.
2. Fix up the Fedora branding... it's not applying as expected, and it's confusing as to why. It's now end of my week and I don't have that solved. It'll have to wait until next week.

Unfortunately, I don't have either of the above in any state that's useful for sharing yet.

But I did share what I have ready to this PR, even if it's not much yet.

garrett:
Committed my current adjustments to make the Cockpit integration modal dialog more PF6-ish in bc1b372fa839652dd8d0f35c6be0ebd718a83d20.

For this to fully work, we also need to adjust the CSS of Cockpit's Storage page when in Anaconda mode.

This can be done by either:

1. setting a class on the `html` element on Cockpit when in Anaconda mode and having custom CSS that's only active when in Anaconda mode
2. having custom CSS in Anaconda that gets injected into the document within the iframe (the Cockpit page) when the page loads, by adding a CSS link to Anaconda; this requires same origin (which I think is fine) and some JS to inject the link to the CSS

The CSS for Cockpit should be something like:

```css
.pf-v6-c-page__main-container {
  border-radius: 0;
  margin: 0;
  border: 0;
}

.pf-v6-c-page__main {
  padding-inline: var(--pf-t--global--spacer--gap--group-to-group--vertical--default);
}

.pf-v6-c-page {
  --pf-v6-c-page--BackgroundColor: var(--pf-t--global--background--color--primary--default);
}

.pf-v6-c-card {
  --pf-v6-c-card--BorderColor: transparent;
}
```

 (If in Cockpit itself, it would need to be in something like `anaconda-mode` as a class on the `html` element, and all of this would have to be then nested in `.anaconda-mode { }`. Otherwise, it would have to be in a file that is meant to be injected in the page.)
 
In today's Anaconda stantup call, @KKoukiou said she would rather have added in Anaconda mode in Cockpit.

KKoukiou:
The LVMonRAID is failing also in other PRs. https://cockpit-logs.us-east-1.linodeobjects.com/pull-800-6f62bb92-20250429-181910-fedora-rawhide-boot-expensive/log.html#7-2

The testDeletePartition should be fixed by https://github.com/rhinstaller/anaconda-webui/pull/801

PR #739, opened 2025-03-28
mount-point-assignment: support disks as target devices

KKoukiou:
Follow up to: 26d0ef70a3e5cc9ace7d134ceb32fd655a7ab56e

We fixed disk target when using cockpit-storage, but we did not remove the limitation from mount point assignment.

Resolves: rhbz#2354044
Relevant to: rhbz#2353524


rvykydal (approved):
Looks good to me.

PR #742, opened 2025-03-31
Do not offer home reuse if there is no GPT disk

rvykydal:
Related: rhbz#2353002

### DIFF
--- a/src/components/storage/scenarios/ReinstallFedora.jsx
+++ b/src/components/storage/scenarios/ReinstallFedora.jsx
@@ -56,6 +56,16 @@ const checkHomeReuse = ({ autopartScheme, devices, originalExistingSystems, sele
         return unknownMountPoints;
     };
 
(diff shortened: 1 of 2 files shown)


KKoukiou (approved):
Thanks for the new test.

rvykydal:
@KKoukiou I didn't realize that when checking home reuse  we need also to pretend BTRFS default scheme. Will add it to the new test when the tests finish (without that the test would not fail on current F42 as it would inhibit Reinstall on LVM default scheme required).

KKoukiou:
> @KKoukiou I didn't realize that when checking home reuse we need also to pretend BTRFS default scheme. Will add it to the new test when the tests finish (without that the test would not fail on current F42 as it would inhibit Reinstall on LVM default scheme required).

True good catch :)

rvykydal:
I think the PR is ready to merge now, apart from the test failures which seem unrelated to me.

PR #745, opened 2025-04-01
cockpit-storage-integration: move storage check steps to custom hook with step handling

KKoukiou:


rvykydal (approved):
Looks good and better to me.

PR #746, opened 2025-04-01
storage: disable home reuse for existing non-linux systems

rvykydal:


rvykydal:
Some home reuse tests are failing
https://cockpit-logs.us-east-1.linodeobjects.com/pull-746-86761907-20250401-145901-fedora-42-boot-storage/log.html
Needs more thinking and maybe update of tests (or patch to be more specific).

rvykydal:
> Some home reuse tests are failing https://cockpit-logs.us-east-1.linodeobjects.com/pull-746-86761907-20250401-145901-fedora-42-boot-storage/log.html Needs more thinking and maybe update of tests (or patch to be more specific).

The patch looks for all the systems including non-selected disks. Up to now we were considering only selected disks for the constraints (single Fedora system etc). I wonder if we should do that also for non-linux systems (take into account only selected disks). I'd be inclined to do so. If so we'd ideally also need a test case having 2 systems on single disk, which may be a bit of challenge to mock (OTOH we don't have it yet for 2 Fedora systems guard)

rvykydal:
Updated to take only selected disks into account.

src/components/storage/scenarios/ReinstallFedora.jsx
```diff
@@ -77,6 +77,14 @@ const checkHomeReuse = ({ autopartScheme, devices, originalExistingSystems, sele
         }
     }
 
+    // Check that no other than Linux system (Windows, MacOS) is found
+    const allSystems = originalExistingSystems.filter(osdata => isCompleteOSOnDisks(osdata, selectedDisks));
+    if (allSystems > linuxSystems) {
```
KKoukiou, line 82: heh I did not know this works.

KKoukiou (approved):

src/components/storage/scenarios/ReinstallFedora.jsx
rvykydal: Oh, neither did I, to be honest.

rvykydal: Fixing

rvykydal:
The commits were added in https://github.com/rhinstaller/anaconda-webui/pull/768

PR #748, opened 2025-04-02
tests: unlock root account and start sshd in interactive defaults

KKoukiou:
`setup` package introduce a change in rawhide, making the root account locked for security purposes by default in installation images.

For enabling our test suite to connect to the booted image this commit:

- Added a %pre section to unlock the root account
- Ensured sshd service starts

Reference: https://src.fedoraproject.org/rpms/setup/c/7ced36d60b67c9e74f7951123225200597e3d2fa


martinpitt (approved):
Thanks for figuring this out!

PR #749, opened 2025-04-02
scenarios: code cleanup: convert availability checker methods to custom hooks

KKoukiou:


src/components/storage/InstallationScenario.jsx
```diff
@@ -73,60 +65,15 @@ const InstallationScenarioSelector = ({
...
-    const diskFreeSpace = useDiskFreeSpace({ devices, selectedDisks });
-    const mountPointConstraints = useMountPointConstraints({ devices, selectedDisks });
-    const usablePartitions = useUsablePartitions({ devices, selectedDisks });
-    const requiredSize = useRequiredSize();
     const { storageScenarioId } = useContext(StorageContext);
-    const originalExistingSystems = useOriginalExistingSystems();
-    const { defaultScheme } = useContext(StorageDefaultsContext);
-
-    useEffect(() => {
-        if ([diskTotalSpace, diskFreeSpace, mountPointConstraints, requiredSize, usablePartitions].some(itm => itm === undefined)) {
```
rvykydal, line 92: Some of this checks got into `useAvailabilityMountPointMapping` hook:
```
        if ([usablePartitions, mountPointConstraints].some(data => data === undefined)) {
            return;
        }
```
in the PR. Other scenario hooks don't seem to have any. Not sure, maybe there need not be any but rather asking to be sure.

rvykydal (approved):
Looks good to me, just wondering about the `undefined` check - see the comment.

src/components/storage/InstallationScenario.jsx
KKoukiou: I added all relevant checks in all scenario availability hooks. Thanks for catching this. :) 

For the record this was only working because JS will allow comparison of undefined with number.

``` js
>> undefined > 0
<- false
>> undefined < 0
<- false
```

rvykydal:
Looks good to me.

PR #752, opened 2025-04-03
storage: mount point mapping: fix wrapping of the custom mount points rows

KKoukiou:
Before:

127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html

Now:

127 0 0 2_9091_cockpit_@localhost_anaconda-webui_index html (1)


garrett:
This is better, but still weird. Why does trash need so much space, for example?

It's supposed to look more like this:

Wizard, partition assignment, inverted, custom mount, with tooltip

garrett (approved):
This is an improvement, but the table should also not waste so much space on the trash action.

The trash action should have `.pf-v5-c-table__action` or `<Td isActionCell>`, like in https://v5-archive.patternfly.org/components/table/#actions

Approving, as it's definitely an improvement, but it doesn't fix the problem, which is partially caused by the trash action having so much space in its cell.

PR #753, opened 2025-04-03
cockpit-storage-integration: prevent possible race condition with selected disk

KKoukiou:


KKoukiou:
I was not able to reproduce the blocker bug still, I am trying to make assumptions what's wrong.

rvykydal (approved):
Looks good to me.

PR #758, opened 2025-04-04
storage: avoid rendering wizard pages while Cockpit Storage is active

KKoukiou:
Rendering wizard pages in the background can trigger API calls to devices that may no longer exist, leading to errors.
We currently work around this by handling and silencing these errors [1] as a hotfix for Fedora 42, due to limited time to implement a proper fix.

This change takes first steps to a more robust approach by preventing the rendering of wizard pages entirely while cockpit-storage mode is active, avoiding invalid device accesses altogether.

[1] https://github.com/rhinstaller/anaconda-webui/commit/53f2ff

### DIFF
--- a/src/components/AnacondaPage.jsx
+++ b/src/components/AnacondaPage.jsx
@@ -28,6 +28,7 @@ export const AnacondaPage = ({
     isFirstScreen,
     isFormDisabled,
     setIsFormDisabled,
+    showStorage,
     step,
     title,
     usePageInit,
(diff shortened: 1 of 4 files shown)


rvykydal (approved):
Looks good to me.
Do we also want to remove the work around? Or would it be done later on?

KKoukiou:
> Looks good to me. Do we also want to remove the work around? Or would it be done later on?

Removing the workaround needs more work. I was experimenting in https://github.com/rhinstaller/anaconda-webui/pull/754 for that, I might get back to that next week.

PR #759, opened 2025-04-04
storage: fix selected disk handling when replacing an MD RAID device

KKoukiou:
When an MD RAID device is deleted and a new one is immediately created using the same disks, the selected disks list will temporarily not include the parent disks. To handle this correctly, we first re-add the parent disks to the selection, then determine whether the new MD RAID device should replace them in the selected set.

Resolves: rhbz#2354798


rvykydal (approved):

KKoukiou:
Clean rebase to make the tests happy.

KKoukiou:
Both test failures are related to some setup logic using rsync, unrelated.

PR #765, opened 2025-04-07
Override globally modals min-width to stretch according to the contents

KKoukiou:
Resolves: rhbz#2357172


garrett (approved):
Thanks! This looks good. :+1:

PR #766, opened 2025-04-07
cockpit-storage-integration: prevent entering the storage editor without selected disks

KKoukiou:
Disks must be selected before entering the storage editor. Currently, entering without a selection leads to unexpected behavior and warnings, as this case isn't properly handled.


KKoukiou:
Do not merge for fedora-42, as it introduces new strings for translation.

KKoukiou:
not very interesting scenario. Let's ignore for now.

PR #767, opened 2025-04-07
mount-point-assignment: ensure partitions on RAID disks are displayed in mount point assignment

KKoukiou:


rvykydal:
The first 6 commits look good to me, I am looking more at the last one.
Or should I wait for tests to pass?

src/helpers/storage.js
```diff
@@ -219,3 +232,50 @@ export const getUsableDevicesManualPartitioning = ({ devices, selectedDisks }) =
 
     return usableDevices;
 };
+
+const reuseManualPartitioning = ({ devices, partitioning, selectedDisks }) => {
+    const usableDevices = getUsableDevicesManualPartitioning({ devices, selectedDisks });
+
+    // Disk devices are not allowed in the mount point assignment
+    const usedDevices = (partitioning?.requests?.map(r => r["device-spec"]) || []);
+    if (usedDevices.every(d => usableDevices.includes(d)) && usableDevices.every(d => usedDevices.includes(d))) {
```
rvykydal, line 241: Maybe we could use checkIfArraysAreEqual here but it is unrelated to the PR.

src/components/storage/InstallationMethod.jsx
```diff
@@ -120,45 +124,42 @@ const CustomFooter = ({ isFormDisabled, isReclaimSpaceCheckboxChecked, setStepNo
...
-            setNewPartitioning(partitioning.path);
+        const part = await getNewPartitioning({
+            autopartScheme: defaultScheme,
+            currentPartitioning: partitioning,
+            devices,
+            method,
+            selectedDisks,
+            storageScenarioId,
+        });
+        setNewPartitioning(part);
```
rvykydal, line 135: Seemed to me that in case of method MANUAL partitioning is passed to setNewPartitioning here, while before it was partitioning.path ?

KKoukiou: I think this is not longer relevant.

src/helpers/storage.js
KKoukiou: used

PR #772, opened 2025-04-10
tests: adjust testMBRParttable to not use rsync

KKoukiou:
Use cp command to copy only what is really necessary for detecting a system. Do not rsync the whole image, this is just very flaky. Also make the affected test nondestructive as they should not fail now.

### DIFF
--- a/test/check-storage-mountpoints
+++ b/test/check-storage-mountpoints
@@ -280,6 +280,9 @@ class TestStorageMountPoints(VirtInstallMachineCase, StorageCase):
         # Create standard btrfs layout in the empty disk and copy the data from the fedora
         # disk to emulate the installation on MBR disk.
         # Then eject the fedora disk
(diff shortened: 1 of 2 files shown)


rvykydal (commented):
IIRC, when I wanted to do similar for testing with kickstart tests, I was hitting - with backend - "arch" tool missing on the system when detecting existing OS:
https://github.com/rhinstaller/anaconda/blob/93af977b0c106390cdd9ef3a2d4b887cdb7b13e4/pyanaconda/modules/storage/devicetree/root.py#L162
Not sure if it is / became less of the issue. The changed test is passing, but maybe it is because it didn't get to calling the function in the backend. It is possible that when we use the same mocking we will hit it later in some other test.

rvykydal:
If we mock the system this way (which would be great) we don't even need the image to copy from maybe (can be too expensive resource), just a library of the mocked files/systems.

PR #776, opened 2025-04-14
build: pin down cockpit-storaged version because of Patternfly 6 changes

KKoukiou:
Unpin when we are also ready with Patternfly 6 migration.

### DIFF
--- a/packaging/anaconda-webui.spec.in
+++ b/packaging/anaconda-webui.spec.in
@@ -20,11 +20,12 @@ BuildRequires: systemd-rpm-macros
 %endif
 
 %global cockpitver 275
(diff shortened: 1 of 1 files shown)


garrett (approved):
Thanks! I approve of pinning Cockpit down.

This should give us a few extra days to make sure everything's ready enough.

(For example: I'm concentrating on Cockpit still, but need to review some of our changes, like the Fedora theme, which currently overrides the blue. Pinning this down means I can finish up the Cockpit work and concentrate on the Anaconda changes without having to rush on both at the same time.

I'm sure it will also free up a bit of time for others too.)

PR #777, opened 2025-04-14
tests: increase the size of the disk image in testDeletePartition

KKoukiou:
New payload is bigger, we need more space. [1]

[1] https://github.com/cockpit-project/bots/pull/7645

### DIFF
--- a/test/check-storage-reclaim
+++ b/test/check-storage-reclaim
@@ -343,7 +343,6 @@ class TestStorageReclaimExistingSystemFedora(VirtInstallMachineCase):
         i.open()
         i.reach(i.steps.INSTALLATION_METHOD)
         s.set_scenario("use-free-space")
-        s.reclaim_set_checkbox(True)
         i.next(True)
 
         b.assert_pixels(


test/check-storage-reclaim
```diff
@@ -343,7 +343,6 @@ class TestStorageReclaimExistingSystemFedora(VirtInstallMachineCase):
         i.open()
         i.reach(i.steps.INSTALLATION_METHOD)
         s.set_scenario("use-free-space")
-        s.reclaim_set_checkbox(True)
```
martinpitt, line 346: It's hard for me to review that -- How does that checkbox relate to increasing the disk image size?

KKoukiou: The checkbox is automatically set to checked and is disabled, is there is not enough space for the new installation on the disk. This payload image refresh, makes the new installation require a few more MB than the previous payload, making the free space on the image not enough for the new installation, therefore the checkbox is now disabled.

martinpitt (approved):
Ack, thanks!

PR #778, opened 2025-04-15
tests: deflake checks using rsync 

KKoukiou:


test/check-storage-mountpoints-btrfs
```diff
@@ -159,11 +159,9 @@ class TestStorageMountPointsBtrfs (VirtInstallMachineCase):
         disk = "/dev/vda"
         tmp_mount = "/tmp/btrfs-mount-test"
         s.partition_disk(disk, [("1MiB", "biosboot"), ("1GB", "ext4"), ("", "btrfs")])
+        s.create_luks_partition(f"{disk}3", "einszweidrei", "encrypted-vol", "btrfs")
         m.execute(f"""
-        # encrypt the btrfs partition
-        echo einszweidrei | cryptsetup luksFormat {disk}3
         echo einszweidrei | cryptsetup luksOpen {disk}3 encrypted-vol
```
rvykydal, line 164: Is this line left intentionally? Shouldn't it be removed?

test/helpers/utils.py
```diff
@@ -43,19 +43,26 @@ def pretend_default_scheme(test, scheme):
...
+                --exclude=/var/lib/machines \
+                --exclude=/var \
+                {source_mountpoint}/* {target_mountpoint}
+        """)
+    except RuntimeError as e:
+        print(f"Error during rsync: {e}")
+        if retry:
+            rsync_directory(machine, source_mountpoint, target_mountpoint, retry=False)
+        else:
+            raise RuntimeError(f"Rsync failed after retrying: {e}") from e
```
rvykydal, line 68: This message might get problematic.
If one calls rsync_directory(... retry=False) from elsewhere right away it can be misleading.

test/check-storage-mountpoints-btrfs
KKoukiou: Actually the `create_luks_device` helper by default closes the device  after creation, we need to have it open for manipulating it. 
however I expose `close_luks` parameter, I will utilize here :) thanks for spotting.

rvykydal (approved):
Looks great to me, I just have two comments/questions. Thank you.

KKoukiou:
> Looks great to me, I just have two comments/questions. Thank you.

Fixed both, thanks

PR #789, opened 2025-04-22
Validate storage show warnings

adamkankovsky:


garrett:
@adamkankovsky: Critical question that would inform design decisions:

Are the recommendations actually recommendations... or requirements? That is, can you successfully ignore them and have a fully working system after installation, or will you hit problems during or after installation if you ignore them? Also: Is the answer the same for all of the "warnings" in the screenshot above, or are some different from the others?

I have some ideas (with a few different paths) on how to address the UX/UI issues, but I need to know the answers to the above questions to proceed. Thanks!

garrett:
I'm still waiting on the questions I asked at https://github.com/rhinstaller/anaconda-webui/pull/789#issuecomment-2958836929

Restated, as a list, with different words:

1. For each one of the points in the screenshot: is it just a _recommendation_ or actually a _requirement_ for things to work?
2. What is the specifically happens if each point is ignored it (for example: the install fails, the system breaks when trying to use it, the system fails to update, etc.)?

Please go through all of the possible points (warnings, errors, notices) — those in the screenshot above, and also others that could show up (not necessarily in the screenshot, in case there are more — and answer the above questions for each please.

**_Also_**, important to note: The layout _cannot_ be ✅ valid if there are errors, and arguably warnings too.

---

I'm waiting for the answers to be able to proceed with recommendations and mockups. I can't proceed until these questions are answered. Thanks!

adamkankovsky:
Everything shown in the screenshot represents recommendations, not hard requirements. These warnings indicate potential issues that won't block installation, but may lead to problems later — such as during system updates, boot, or general usage. While you can proceed with installation despite these warnings, doing so may result in:
- Failures during system updates (e.g. if /boot is too small),
- Boot issues (e.g. due to missing or reused EFI partition),
- Limited system usability (e.g. if /home is too small or no swap is defined).
On the other hand, errors must be resolved before installation can proceed — they are not optional. Even if the system installs successfully with warnings, it may later encounter critical issues like failed kernel updates, insufficient disk space, or the inability to hibernate.

garrett:
OK, I had the start of a mockup, but realized that the messages needed overhaul, so I asked you for feedback.

Then I thought I posted a comment here with some revisions. But I don't see it, so here is another instead (which is probably pretty close to the original one I am sure I typed):

Here are revised messages, in the same order as your last comment, attempting to more clearly communicate each message. The very first sub point is the type of alert (some errors should be warnings and vice versa, IIRC?).

I tried to make it so that the new versions concisely describe the problem and mention a solution.

1.  `It is recommended to create a new file system on your %(mount)s partition.`
    
    * _Warning_
    * Format `$%(mount)s` to avoid issues with existing data.
    * _When there are multiples:_ Format these partitions to avoid issues with existing data: `/boot`, `/home`.
    - ...Or, alternate for multiples: Re-use singular and change the mount substring to a sentence list. In English, it'd be like: `Format /boot and /home to avoid issues with existing data.` Or `Format /boot, /efi, and /home to avoid issues with existing data.` ... it would need to be localized per language. I hope there's a function that can be used for that.

2.  `Your %(mount)s partition is less than %(size)s which is lower than recommended for a normal %(productName)s install.`
    
    * _Warning_
    * `$%(mount)s` should be at least `$%(size)s`.

3.  `Any warning from storage.bootloader.warnings (from verify_bootloader – these are dynamic and can vary)`
    
    * _Error_
    * Bootloader error: `[Error text goes here]`

4.  `You have not specified a swap partition. %(requiredMem)s of memory is recommended to continue installation without a swap partition, but you only have %(installedMem)s.`
    
    * _Error_
    * Installation requires `$%(requiredMem)s` of RAM, but only `$%(installedMem)s` is available. Create a swap partition.

(I think this is a warning, but should be an error?)

5.  `A swap partition has not been specified. To significantly improve performance for most installations, it is recommended to specify a swap partition.`
    
    * _Note_
    * Only `$%(size)s` of memory is available. Create a swap partition for a faster installation experience.

6.  `At least one of your swap devices does not have a UUID, which is common in swap space created using older versions of mkswap. These devices will be referred to by device path in /etc/fstab, which is not ideal since device paths can change under a variety of circumstances.`
    
    * _Warning_
    * Swap is referred to by its device path and might not be found during boots. Recreate the swap partition and assign it a stable UUID.

7.  `The available memory is less than %(size)s which can be too small for LUKS2 format. It may fail.`
    
    * _Warning_
    * LUKS2 disk encryption might fail, as less than `$%(size)s` of RAM is available. Creating a swap partition may help.

Are these all the possible errors we'd encounter in this screen? I can try to integrate them into a mockup, if so.

garrett:
OK, wow. That's a lot of messages! I've rewritten them all, with classification and some notes, but I could be wrong... there might be nuance in the code or such, so please feel free to point out any issues and it's OK second-guess everything.

I rewrote the messages on your new list (which includes ones from before, with a little more polishing) to abide by these general guidelines for UI messaging:

1. **Be clear**, so it's obvious what is being communicated. This means avoiding jargon, trying to not have filler words (including "please"), and breaking down complex sentences into simpler ones.
2. **Provide a solution**. While errors can sometimes imply what to do next... we don't want anyone stuck, wondering what to do to fix the problem, so we should say what to do to be able to successfully finish installation.
3. **Be specific**. Examples: When minimum and maximum sizes are needed, show what they are. When specific partitions need attention, say which ones. (This builds on 1 and 2 of course, to have additional details to be clearer and help to provide a solution by saying what specifically needs to be done.)
4. **Use neutral language**. Sometimes the old errors blame people ("You did not create a partition"), but we want to avoid that. Errors are not supposed to be a person's fault; they're the result of the computer not doing what someone wants it to do, so we want to phrase them in a non-accusatory way.

I also tried to put it in a problem-first, solution second construction... and I opted to make it active voice as long as it wouldn't make things clunky.

For reference, there are a lot of other guidelines on PatternFly's website:

- https://www.patternfly.org/ux-writing/error-messages
- https://www.patternfly.org/ux-writing/ux-writing-best-practices
- https://www.patternfly.org/ux-writing/terminology

And more on the subject in general @ NNG, which is also good reading / watching:

- https://www.nngroup.com/articles/error-message-guidelines/
- https://www.youtube.com/watch?v=sReni_EeZUM

(Some of these ideas stem from Jakob Nielsen's classic "10 Usability Heuristics for User Interface Design" @ https://www.nngroup.com/articles/ten-usability-heuristics/ )

The format is the same as before, with a numbered list per item with the original string, with a sub-list with the type, where it's from (_new_), any notes for you (as "developer notes"... also _new_), and it ends with the revised string.

---

### 1. `It is recommended to create a new file system on your %(mount)s partition.`

* **Type**: Warning
* **From**: `verify_partition_formatting`
* To avoid conflicts with existing data, reformat `%(mount)s`.

### 2. `Your %(mount)s partition is less than %(size)s which is lower than recommended for a normal %(productName)s install.`

* **Type**: Warning
* **From**: `verify_partition_sizes`
* The `%(mount)s` partition should be at least `%(size)s` for a standard installation.

### 3. `Your %(mount)s partition size is lower than required %(size)s.`

* **Type**: Error
* **From**: `verify_partition_sizes`
* The `%(mount)s` partition must be at least `%(size)s`.

### 4. `Your %(mount)s partition is too small for %(format)s formatting (allowable size is %(minSize)s to %(maxSize)s).`

* **Type**: Error
* **From**: `verify_partition_format_sizes`
* The `%(mount)s` partition is too small for `%(format)s`. It must be between `%(minSize)s` and `%(maxSize)s`.

### 5. `Your %(mount)s partition is too large for %(format)s formatting (allowable size is %(minSize)s to %(maxSize)s).`

* **Type**: Error
* **From**: `verify_partition_format_sizes`
* The `%(mount)s` partition is too large for `%(format)s`. It must be between `%(minSize)s` and `%(maxSize)s`.

### 6. `No valid boot loader target device found. See below for details.`

* **Type**: Error
* **From**: `verify_bootloader`
* **Developer note**: I'm not sure what the "details" below this would be. If there are details, we can just list it without having to say "see below for details"?
* A valid target for the boot loader was not found. Create a `/boot` partition to continue.

### 7. `(Platform stage1 suggestion – dynamic text via platform.stage1_suggestion.)`

* **Type**: Error
* **From**: `verify_bootloader`
* Boot loader error: `[Dynamic error text appears here]`.

### 8. `You have not created a bootable partition.`

* **Type**: Error
* **From**: `verify_bootloader`
* No partition is marked as "bootable". Set the boot flag on the appropriate partition (either on `/` or a `/boot` partition).

### 9. `Any error from storage.bootloader.errors (dynamic and can vary).`

* **Type**: Error
* **From**: `verify_bootloader`
* Boot loader error: `[Dynamic error text appears here]`.

### 10. `Any warning from storage.bootloader.warnings (dynamic and can vary).`

* **Type**: Warning
* **From**: `verify_bootloader`
* Boot loader warning: `[Dynamic warning text appears here]`.

### 11. `Any error from storage.bootloader.check() (dynamic and can vary).`

* **Type**: Error
* **From**: `verify_bootloader`
* Boot loader error: `[Dynamic error text appears here]`.

### 12. `You have not defined a root partition (/), which is required for installation of %(productName)s to continue.`

* **Type**: Error
* **From**: `verify_root`
* A root partition (/) is required for installation. Create one to continue.

### 13. `You must create a new file system on the root device.`

* **Type**: Error
* **From**: `verify_root`
* The root partition (/) must be reformatted to continue.

### 14. `Your root partition must be on a device of type: %(types)s.`

* **Type**: Error
* **From**: `verify_root`
* The root partition must be on one of the following device types: `%(types)s`.

### 15. `The LDL DASD disk %(disk)s (%(busid)s) cannot be used for the installation. Please format it.`

* **Type**: Error
* **From**: `verify_s390_constraints`
* The LDL DASD disk `%(disk)s` is unusable. Format it to the CDL layout to continue.

### 16. `Your BIOS-based system needs a special partition to boot from a GPT disk label. To continue, please create a 1MiB 'biosboot' type partition on the %(disk)s disk.`

* **Type**: Error
* **From**: `verify_gpt_biosboot`
* Booting from a GPT disk on a BIOS system requires a 1 MiB 'biosboot' type partition on `%(disk)s`. Create one to continue.

### 17. `The system will not be bootable. The firmware does not support XFS file system features on the boot file system. Upgrade the firmware or change the file system type.`

* **Type**: Error
* **From**: `verify_opal_compatibility`
* System firmware does not support booting from an XFS partition. To continue, select a different file system type for the boot partition or upgrade the firmware.

### 18. `You have not specified a swap partition. %(requiredMem)s of memory is recommended to continue installation without a swap partition, but you only have %(installedMem)s.`

* **Type**: Warning
* **From**: `verify_swap`
* **Developer note**: We can calculate `swapSize` here as the difference between `%(requiredMem)s` and `%(installedMem)s`. I'm not sure if we should use requiredMem - installedMem or just use installedMem as the minimum size. Or, perhaps, even installedMem x 2, as that was the old "rule of thumb" for systems with low memory years ago. (Which this message is about.)
* The system has `%(installedMem)s` of memory, but `%(requiredMem)s` is recommended. For better performance, create a swap partition of at least `%(swapSize)s`.

### 19. `You have not specified a swap partition. %(requiredMem)s of memory is required to continue installation without a swap partition, but you only have %(installedMem)s.`

* **Type**: Error
* **From**: `verify_swap`
* **Developer note**: See the entry above (number 18) about `swapSize`.
* Insufficient memory to install. `%(installedMem)s` is available, but `%(requiredMem)s` is required. Create a swap partition of at least `%(swapSize)s`.

### 20. `A swap partition has not been specified. To significantly improve performance for most installations, it is recommended to specify a swap partition.`

* **Type**: Note
* **From**: `verify_swap`
* This system only has %(size)s of available memory. Creating a swap partition would dramatically improve system reliability in most scenarios.

### 21. `At least one of your swap devices does not have a UUID... These devices will be referred to by device path in /etc/fstab, which is not ideal...`

* **Type**: Warning
* **From**: `verify_swap_uuid`
* A swap partition is using a device path that can change at boot, which might prevent the system from finding it. Recreate the partition to assign a stable UUID.

### 22. `The mount point %(mountpoint)s must be on a linux file system.`

* **Type**: Error
* **From**: `verify_mountpoints_on_linuxfs`
* **Developer note**: "Linux" should be capitalized.
* The mount point `%(mountpoint)s` must be on a Linux file system.

### 23. `This mount point is invalid. The %(dir)s directory must be on the / file system.`

* **Type**: Error
* **From**: `verify_mountpoints_on_root`
* The `%(dir)s` directory must be on the root (/) file system, not a separate partition.

### 24. `Your %(dir)s must be on a separate partition or LV.`

* **Type**: Error
* **From**: `verify_mountpoints_not_on_root`
* The `%(dir)s` directory must be on its own separate partition or logical volume.

### 25. `The existing unlocked LUKS device %(dev)s cannot be used for the installation without an encryption key specified for this device. Please, rescan the storage.`

* **Type**: Error
* **From**: `verify_unlocked_devices_have_key`
* The unlocked LUKS device `%(dev)s` requires an encryption key. Rescan storage and provide its key to continue.

### 26. `Encryption requested for LUKS device %(dev)s but no encryption key specified for this device.`

* **Type**: Error
* **From**: `verify_luks_devices_have_key`
* Encryption for `%(dev)s` requires a passphrase or encryption key. Enter one to proceed.

### 27. `The available memory is less than %(size)s which can be too small for LUKS2 format. It may fail.`

* **Type**: Warning
* **From**: `verify_luks2_memory_requirements`
* LUKS2 disk encryption might fail with less than `%(size)s` of memory. Creating a swap partition may help.

### 28. `%(path)s is currently mounted and cannot be used for the installation. Please unmount it and retry.`

* **Type**: Error
* **From**: `verify_mounted_partitions`
* The partition `%(path)s` is currently in use. Unmount the partition to use it for installation.

### 29. `Selected disks %(disks)s contain volume group '%(vg)s' that also uses further unselected disks. You must select or de-select all these disks as a set.`

* **Type**: Error
* **From**: `verify_lvm_destruction`
* **Developer note**: `blivet` should be able to get the full list of disks in the VG, right? Ideally, we could use this to find the missing disks and then offer to auto-select them, instead of making the person installing have to guess. Meanwhile, I've rewritten the message so we can do an in-place improvement without extra work, for now.
* The volume group `%(vg)s` spans multiple disks, but the current selection only includes `%(disks)s`. Select or deselect all disks in this group to continue.

garrett:
Thanks for the update! I don't have time right now to properly review and provide feedback, but my initial thoughts are these:

1. Not enough space between the table and the alert. (It needs just a little more. Not sure which spacer. Also not sure why it isn't spaced by default.)
2. The new labels here help a lot: For example, "The following partitions should be at least 100 MiB" is silly when it's there 2x and each has 1 item. I'll help with rewording. We either can combine them (but the size are different, so probably not), or reword it a bit (which I think can get smaller and more direct).

I intend on helping out with # 2 when I get some time. Thanks! (Probably next week, unless I find a few minutes soon.)

garrett:
If there is a warning (or are warnings), with an error too, it would look like this (on the right):

Group

KKoukiou:
Let's close the PR and possibly reopen once this is being worked on again.

PR #799, opened 2025-04-29
Review screen bootloader location

adamkankovsky:
image
after cockpit storage:
image
efi after cockpit storage:
image

### DIFF
--- a/src/components/review/StorageReview.jsx
+++ b/src/components/review/StorageReview.jsx
@@ -29,6 +29,7 @@ import {
     getDeviceChildren,
     getParentPartitions,
     hasEncryptedAncestor,
+    isBootloaderDevice,
 } from "../../helpers/storage.js";
 
 import {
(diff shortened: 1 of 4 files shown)


KKoukiou (commented):
Also tests seem quite broken.

KKoukiou (commented):
from the screenshot I see that bootloader is appearing in the second column, I would like to see:
* format as efi / biosboot (if the bootloader is being reformatted)
* mount efi/biosboot (if we are reusing the existing bootloader)

Basically treat efi/biosboot as other format types.

What do you think? 

Basically it's important to know if we are reformatting the bootloader partition or not. 
If this information is not available by the existing APIs, feel free to land this and we folow up with another iteration.

KKoukiou:
Please rebase for tests re-run.

KKoukiou:
I moved the logic to a helper function, I will send a follow up for the 'format as ...[bootloader type]'

KKoukiou:
Updated the pixel tests and rebased.

KKoukiou (approved):

PR #802, opened 2025-04-30
Anaconda needs latest pykickstart but image refresh is currently blocked on kernel regression

KKoukiou:
This should be reverted ASAP.


jkonecny12 (approved):
LGTM!

PR #807, opened 2025-05-07
branding: Simplify Fedora colors

garrett:
Only override branding colors, based on Fedora; this simplifies the branding CSS a good bit and has a similar effect.


garrett:
(Aha, GitHub Copilot actually caught both of the issues, but Sourcery only caught 1 of them. :wink: Re-pushing with the actual correct CSS now.)

KKoukiou (approved):

PR #810, opened 2025-05-08
apis: remove workaround for previously missing API

KKoukiou:


adamkankovsky (approved):
LGTM, This is no longer necessary.

PR #820, opened 2025-05-13
components: storage: fix disabled radio buttons

KKoukiou:
Patternfly 6 does not expose isAriaDisabled property on radio buttons. https://www.patternfly.org/components/forms/radio/


adamkankovsky (approved):
Thanks for the fix

rvykydal (approved):

PR #821, opened 2025-05-13
docs: add release guidelines

KKoukiou:
https://github.com/KKoukiou/anaconda-webui/blob/main/docs/release.rst

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -1,0 +1,97 @@
+=================
+Release Process
+=================
+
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me. However, I wonder why we don't have automatic releases. So many manual steps seems unnecessary to me.

KKoukiou:
> Looks good to me. However, I wonder why we don't have automatic releases. So many manual steps seems unnecessary to me.

We need to create a script like makebumpver for Web UI repository. We just don't yet have it :)

PR #824, opened 2025-05-14
tests: Add structured docstrings to Web UI tests

KKoukiou:
Added docstrings to several test cases using the standard format: Description, Setup, Test Steps, and Expected Results. Improves readability and test documentation consistency.

### DIFF
--- a/test/check-storage-cockpit
+++ b/test/check-storage-cockpit
@@ -52,8 +52,14 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testBasicLiveISO(self):
-        # Test that the 'Modify storage' button is available on the first page
-        # https://bugzilla.redhat.com/show_bug.cgi?id=2336488
+        """
+        Description:
+            Test that the 'Modify storage' button is available on the first page
+            of the installer when using a live ISO.
+
+        References:
+            - https://bugzilla.redhat.com/show_bug.cgi?id=2336488
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -68,6 +74,18 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testEncryptedUnlock(self):
+        """
+        Description:
+            Test installation on an encrypted partition created in Cockpit storage
+
+        Expected results:
+            - An encrypted partition can be created and used as the root mount point
+            - An encrypted partition can be reformatted in Cockpit storage and reused
+            - The partition can be deleted in Cockpit storage
+
+        References:
+            - https://bugzilla.redhat.com/show_bug.cgi?id=2354497
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -131,6 +149,13 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testLVM(self):
+        """
+        Description:
+            Test installation on an LVM volume created in Cockpit storage
+
+        Expected results:
+            - An LVM volume can be created and used for the root mount point
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -248,15 +273,38 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testStandardPartitionExt4(self):
+        """
+        Description:
+            Test installation on an ext4 partition created in Cockpit storage
+
+        Expected results:
+            - An ext4 partition can be created and used for the root mount point
+        """
         self._testStandardPartition("ext4")
 
     @nondestructive
     def testStandardPartitionXFS(self):
+        """
+        Description:
+            Test installation on an XFS partition created in Cockpit storage"
+
+        Expected results:
+            - An xfs partition can be created and used for the root mount point
+        """
         self._testStandardPartition("xfs")
 
     @nondestructive
     @run_boot("efi")
     def testBtrfsTopLevelVolume(self):
+        """
+        Description:
+            Test installation on a btrfs top-level volume created in Cockpit storage
+
+        Expected results:
+            - A btrfs top-level volume can be created and used for the root mount point
+            - The review screen shows the correct mount points
+            - The fstab file is updated correctly
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -296,6 +344,15 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testBtrfsSubvolumes(self):
+        """
+        Description:
+            Test installation on a btrfs subvolume created in Cockpit storage
+
+        Expected results:
+            - A btrfs subvolume can be created and used for the root and home mount points
+            - The review screen shows the correct mount points
+            - The fstab file is updated correctly
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -347,6 +404,16 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testBtrfsAndMountPointAssignment(self):
+        """
+        Description:
+            Test installation on a btrfs subvolume created in Cockpit storage by
+            selecting these in the Mount Point Assignment screen
+
+        Expected results:
+            - A btrfs subvolume can be created and used through mount point assignment screen for the root and home mount points
+            - The review screen shows the correct mount points
+            - The fstab file is updated correctly
+        """
         b = self.browser
         m = self.machine
         i = Installer(b, m, scenario="use-configured-storage")
@@ -415,13 +482,15 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
     @nondestructive
     def testRAIDUnsupportedNesting(self):
         """
-        Check #1: Verify that partitions created on a RAID device which is on top of partitions are considered unsupported
+        Description:
+            Check that partitions created on a RAID device which is on top of partitions are considered unsupported
 
-        This logic is currently implemented in the frontend, but it should be moved to the backend.
+        References:
+            - https://issues.redhat.com/browse/INSTALLER-4141
 
-        See: https://issues.redhat.com/browse/INSTALLER-4141
-
-        Once this is moved to backend we should implement a unit test for this and we can remove the Web UI test.
+        TODO:
+            This logic is currently implemented in the frontend, but it should be moved to the backend
+            Once this is moved to backend we should implement a unit test for this and we can remove the Web UI test
         """
 
         b = self.browser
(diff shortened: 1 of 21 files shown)


KKoukiou:
Let's standardize some docstring template, so that I can apply it to all tests. Let me know about your suggestions. @jstodola @jikortus

jstodola:
It looks good to me in general, thanks.

When looking at a test case, the most important information for me is what the test is supposed to cover, having a link to jira/bugzilla so that it's easier to find out why we decided to test the specific scenario, and what is the expected behavior.

I don't really mind that test steps are missing for automated tests:
- they are automated, so it's unlikely (based on my experience) that the tests would be executed manually. If there is a problem with the test, it would be fixed or disabled (and then fixed), so manual steps are not too important.
- as @jikortus mentioned, sometimes it's not easy to keep the test steps and the actual automation in sync over time (also our own experience).

IMO, instead of having test steps explicitly described, it's better to add more comments to the test code itself, where it makes sense for the future maintenance.

And I agree with removing the empty lines after each section title.

KKoukiou:
@jikortus @jstodola I agree that we should avoid including detailed Test Steps in the docstrings for the reasons @jstodola mentioned above.

Also, most of these tests were written when features were developed as part of the test-driven development approach and aren't directly tied to specific Bugzilla or Jira tickets. Where applicable, I’ve included a References: section to link relevant issues, but in general, those are the exception rather than the rule.

I also removed the extra spaces as requested.

Let me know if any additional changes are needed — happy to adjust!

jikortus (approved):
Looks good to me as well, thank you!

test/check-storage-cockpit
```diff
@@ -68,6 +74,18 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testEncryptedUnlock(self):
+        """
+        Description:
+            Test using for the installationi an encrypted partition created in Cockpit storage
```
rvykydal, line 79: A typo "installationi", the sentence sounds broken.

test/check-storage-cockpit-e2e
```diff
@@ -26,6 +26,19 @@ from testlib import test_main  # pylint: disable=import-error
...
+        """
+        Description:
+            Test scenario with three disks:
+
+            - 'biosboot' and '/boot' are on 'vda'.
+            - A RAID 0 array is created using partitions on 'vdb1' and 'vdc1'.
+            - The root filesystem ('/') is placed on the RAID 0 device.
+
+        Expected results:
+            - The installer can created a RAID0 array on partitions and use it for rootfs
```
rvykydal, line 38: A typo "created"

test/check-storage-home-reuse
```diff
@@ -156,6 +174,16 @@ class TestStorageHomeReuseFedora(VirtInstallMachineCase):
...
     @disk_images([("fedora-rawhide", 15), ("fedora-41", 15)])
     def testHomeReuseFedoraEncrypted(self):
+        """
+        Description:
+            Test the home reuse installation method with a pre-existing standard
+            Fedora installation layout with encrypted /.
+
+        Expected results:
+            - The home reuse installation method is available
+            - The encrypted device needs to be unlocked
```
rvykydal, line 184: I'd squash this into single
`- The home reuse installation method is available after the encrypted device is unlocked.`
I think we could update the test to check that it is not available before the unlocking.
`- The home reuse installation method is available only after the encrypted device is unlocked.`

test/check-storage-home-reuse
```diff
@@ -189,6 +217,19 @@ class TestStorageHomeReuseFedora(VirtInstallMachineCase):
...
+            Test the home reuse installation method with a pre-existing multiple Linux OSes
+            on the selected disks
+
+        Expected results:
+            The home reuse installation method is not available
+                - when two Fedora systems are on the selected disks
+                - when no Fedora system is on the selected disks
+
+            The home reuse installation method is available
+                - when only one Fedora system and possibly other Linux OSes are on the selected disks
```
rvykydal, line 231: This does not reflect the tests (and expected state) correctly.
It should be available only when single Linux OS is available and the OS is Fedora OS.
And it explicitly checks that it is not available in case of
- two Fedora OSes
- single non-Fedora Linux OS is available
- Fedora OS and another Linux OS is available

test/check-storage-mountpoints
```diff
@@ -191,6 +212,14 @@ class TestStorageMountPoints(VirtInstallMachineCase, StorageCase):
     @disk_images([("", 15), ("", 10)])
     @nondestructive
     def testMultipleDisks(self):
+        """
+        Description:
+            Test the case when multiple disks are selected.
```
rvykydal, line 217: The test is also checking boot partition requirement - not sure if we have more specific test for that, but if not it should be perhaps mentioned here.

test/check-storage-mountpoints
```diff
@@ -361,6 +410,18 @@ class TestStorageMountPoints(VirtInstallMachineCase, StorageCase):
 
     @nondestructive
     def testEncryptedUnlockCockpit(self):
+        """
+        Description:
+            Test 'Mount point mapping' with encrypted partitions created using Cockpit Storage
+
+        Expected results:
+            - The user can create an encrypted partition with cockpit storage
+            - The 'Unlock' dialog is displayed before entering the mount point mapping
```
rvykydal, line 419: I'd rather say
`- The 'Unlock' dialog is available for unlocking the partitions before entering the mount point mapping`

test/check-storage-mountpoints-btrfs
```diff
@@ -171,6 +179,18 @@ class TestStorageMountPointsBtrfs (VirtInstallMachineCase):
         """)
 
     def testEncryptedUnlockBTRFSonLUKS(self):
+        """
+        Description:
+            Test that the installer can handle BTRFS on LUKS encrypted devices
+
+        Expected results:
+            - The 'Unlock' dialog is displayed before entering the mount point mapping
+              screen.
```
rvykydal, line 188: I'd rather say
`- The 'Unlock' dialog is available for unlocking the partitions before entering the mount point mapping`

test/check-storage-cockpit
KKoukiou: It's italian installation :P

test/check-storage-reclaim
```diff
@@ -172,6 +188,17 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
         self.setup_partitions(s)
 
     def testReclaimSpaceShrinkBtrfsSubvolumes(self):
+        """
+        Description:
+            Test that the reclaim space scenario with shrinking partitions
```
rvykydal, line 193: A typo - remove 'that'?

test/check-storage-reclaim
```diff
@@ -58,8 +58,15 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
...
 
     def testReclaimSpaceOptional(self):
-        # When there is enough free space for the installation
-        # the reclaim action is not set by default
+        """
+        Description:
+            Test that the reclaim space scenario when there is enough free space
+            for the installation.
+
+        Expected:
```
rvykydal, line 66: s/Expected/Expected results/

test/check-storage-reclaim
```diff
@@ -369,6 +412,13 @@ class TestStorageReclaimExistingSystemWindows(VirtInstallMachineCase):
 
     @disk_images([("", 20)])
     def testBasic(self):
+        """
+        Description:
+            Test reclaiming space from an existing Windows system by deleting partitions
+
+        Expected results:
+            - The review screen shows the affected system when partitions from an existing system are deleted
```
rvykydal, line 420: I think this test (compared to the one above (`testDeletePartition`) is missing the review screen check.

rvykydal (commented):
Looks great to me, just have a few comments.

test/check-storage-home-reuse
KKoukiou: I think I got this now right?

test/check-storage-reclaim
KKoukiou: Did not add the check, because I wanted to keep this commit clean from code changes. Adjusted the docstring, and we can follow up with improvements of the code itself.

test/check-storage-cockpit
```diff
@@ -248,15 +273,38 @@ class TestStorageCockpitIntegration(VirtInstallMachineCase, StorageCase):
...
+        Expected results:
+            - An ext4 partition can be created and used for the root mount point
+        """
         self._testStandardPartition("ext4")
 
     @nondestructive
     def testStandardPartitionXFS(self):
+        """
+        Description:
+            Test using for the installation an xfs partition created in Cockpit storage
```
jikortus, line 289: Could this description be reworded a bit? I get what it means, but it took me a while to figure out :-). What about e. g. "Test installation on an XFS partition created in Cockpit storage"?

jikortus: I have the same suggestion also for the btrfs scenarios below.

test/check-storage-mountpoints-btrfs
```diff
@@ -171,6 +179,17 @@ class TestStorageMountPointsBtrfs (VirtInstallMachineCase):
...
 
     def testEncryptedUnlockBTRFSonLUKS(self):
+        """
+        Description:
+            Test that the installer can handle BTRFS on LUKS encrypted devices
+
+        Expected results:
+            - The 'Unlock' dialog is available for unlocking the partitions before entering the mount point mapping
+            - The user can unlock the disk using a passphrase
+            - The user can selected the unlocked devices in the mount point mapping
```
jikortus, line 189: typo (selected)

jikortus (commented):
I've found a few spots that could possibly use a slight improvement, but overall the doc strings look good. Thank you @KKoukiou!

test/check-storage-reclaim
```diff
@@ -58,8 +58,15 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
         )
 
     def testReclaimSpaceOptional(self):
-        # When there is enough free space for the installation
-        # the reclaim action is not set by default
+        """
+        Description:
+            Test the the reclaim space scenario when there is enough free space
```
rvykydal, line 63: "the the"

test/check-storage-mountpoints
rvykydal: Actually I think the biosboot check should go to the Description as a second line.
`Test the bootloader partition requirement is applied to the option availability`
because it does not logically belong to the current description (multiple disks).
This way we are aware right away that this test is testing also something specific (which would maybe ideally be in a separate, maybe more comprehensive test for the requirements).

rvykydal (approved):
Looks good to me.

test/check-storage-mountpoints
```diff
@@ -191,6 +212,15 @@ class TestStorageMountPoints(VirtInstallMachineCase, StorageCase):
...
     @nondestructive
     def testMultipleDisks(self):
+        """
+        Description:
+            Test the case when multiple disks are selected.
+
+        Expected results:
+            - The mount point mapping scenario is available and can be used with
+              multiple disks.
+            - The bootloader requirements (biosboot) are verified for the selected disk[s]
```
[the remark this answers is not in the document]
KKoukiou: I would prefer the description to be a flowing text, instead of bullet points I think.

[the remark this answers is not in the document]
KKoukiou: Please check the adjustment if you like it.

PR #826, opened 2025-05-14
Sourcery review fix tests

KKoukiou:


test/helpers/users.py
```diff
@@ -107,7 +107,7 @@ def set_valid_root_password(self, valid=True):
         p = Password(self.browser, ROOT_ACCOUNT_ID_PREFIX)
         password = "password"
         p.set_password(password)
-        p.set_password_confirm(password + "" if valid else "X")
+        p.set_password_confirm(f"{password}" if valid else "X")
```
rvykydal, line 110: This is a bit strange, I'd use just password, but I don't care.

rvykydal (changes requested):
Found only the f-string issue. Other than that looks good to me.

test/helpers/users.py
KKoukiou: It was incorrectly fixed, because of the concatenation, which was probably redundant.

KKoukiou:
Pure rebase, and on typo fix. iGB => GiB

PR #829, opened 2025-05-16
Revert "Anaconda needs latest pykickstart but image refresh is currently blocked on kernel regression"

KKoukiou:
This reverts commit b6702ead655b8290ff766d8949c148b6c1bc0bc1.


KKoukiou:
> Looks good to me but can you please add details why the revert is needed?

We were overriding the pykickstart with a newer RPM applied through updates image. We generally want to keep the updates.img as minimal as possible at all time, there is not other reason that this revert is needed. It would keep working with the newer pyckstart as well.

PR #835, opened 2025-05-19
infra: trigger dependabot workflow for updating node_modules

KKoukiou:
The node cache must be handled when updating package.json, which is managed by our custom Dependabot GitHub workflow.
Previously, we weren't setting the [node-modules] label, so the workflow wasn't triggered. This change ensures that the workflow runs as intended.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,9 +12,9 @@ updates:
       timezone: "Europe/Berlin"
     open-pull-requests-limit: 3
     commit-message:
-      prefix: "infra"
+      prefix: "[no-test]"
     labels:
-      - "sanity check required"
+      - "node_modules"
     groups:
       eslint:
         patterns:
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Force push node_modules update
         run: |
           # Dependabot prefixes the commit with [no-test] which we don't want to keep in the commit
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -12,9 +12,9 @@ updates:
       timezone: "Europe/Berlin"
     open-pull-requests-limit: 3
     commit-message:
-      prefix: "infra"
+      prefix: "[no-test]"
```
jkonecny12, line 15: I'm missing an explanation of this change in the commit message. Why are we changing this to `[no-tests]` to change it back in the other workflow. Is the prefix necessary for another automation?

jkonecny12 (approved):
Looks good to me otherwise.

.github/dependabot.yml
KKoukiou: This PR is marked with the no-test label, which tells the Cockpit bots to skip CI testing initially: https://github.com/cockpit-project/bots/blob/main/tests-scan#L163

Here's how the workflow operates:

-     Dependabot (configured via .github/dependabot.yml) opens a PR to update package.json.
-     We also need to regenerate and commit the updated node_modules, which we are storing at https://github.com/rhinstaller/node-cache
-     To avoid running tests on an incomplete PR, we use the no-test label.
-     The presence of the node-modules label then triggers .github/workflows/dependabot.yml, which commits the generated node_modules to the same PR.
-     Once both package.json and node_modules are in place, the PR is ready for testing therefore we remove the 'no-test' label and put the 'infra' expected label.

PR #839, opened 2025-05-22
Use process inspection to detect GNOME instead of XDG_CURRENT_DESKTOP

KKoukiou:
Cockpit's spawn environment does not expose XDG_CURRENT_DESKTOP, so desktop detection via that variable was unreliable. This change uses `ps` to check for the presence of `gnome-shell` to infer if GNOME is running. Falls back to "UNKNOWN" otherwise.


jkonecny12 (approved):
Looks good to me. Thanks for a quick fix.

PR #855, opened 2025-06-03
tests: increase consistency on how we create partitions

KKoukiou:


test/helpers/storage.py
```diff
@@ -281,15 +276,14 @@ def partition_disk(self, disk, partitions_params, is_mbr=False):
...
                     "swap": "82",
                     "lvmpv": "8e",
-                    "ext4": "83",
-                    "btrfs": "83",
-                    "logical": "83",
                     "extended": "extended"
                 }.get(fstype, "83")
             else:
                 # Determine the GPT type name
                 type_code = {
```
rvykydal, line 283: Here we should add also type code for 'efi'. Without it it won't be detected by the backend when detecting Windows OS.
I am doing it in https://github.com/rhinstaller/anaconda-webui/pull/858 here: https://github.com/rvykydal/anaconda-webui/blob/home-reuse-windows/test/helpers/storage.py#L291. I add also 'win-re' which is not important, but I think 'efi' should be added into this patchset.
I'll rebase my PR then.

KKoukiou:
> I think we should add also type code for 'efi' in this PR.

This was added and merged by some other PR in the meantime.

PR #860, opened 2025-06-04
New Date and Time screen

adamkankovsky:


src/components/timezone/DateAndTime.jsx
```diff
@@ -0,0 +1,411 @@
...
+    const [ntpServer, setNtpServer] = useState("");
+    const [ntpServers, setNtpServers] = useState([]);
+    const [date, setDate] = useState("");
+    const [time24, setTime24] = useState("");
+
+    // Validity: either NTP is enabled, or both date & time are valid
+    useEffect(() => {
+        setSectionValid(
+            autoDateTime ||
+            (!!date && !!time24)
```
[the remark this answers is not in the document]
adamkankovsky: But that's not what's happening. Either it is automatically selected, or the time and date are set. There are double negations. It just makes sure it's always a boolean.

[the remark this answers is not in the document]
KKoukiou: I just noticed these are double negations, sorry, blind.

jstodola:
For me it would make more sense to change the order of items on the screen:
- Timezone
- Date and time
- Time format

In case the timezone is detected and set incorrectly, time will also be incorrect. Fixing the timezone will also fix time. For that reason I suggest to place the timezone selection first on the screen.
The timezone selection in the GKT UI is also placed above date/time settings:
GTK_UI_date_time

test/check-timezone
```diff
@@ -0,0 +1,206 @@
...
+        m = self.machine
+        i = Installer(b, m)
+        dt = DateAndTime(b, m)
+
+        i.open()
+        i.reach(i.steps.DATE_TIME)
+        dt.select_ntp_server("2.fedora.pool.ntp.org")
+        dt.check_ntp_server("2.fedora.pool.ntp.org")
+
+    def testManualTimezone(self):
```
KKoukiou, line 102: Can we do some test merges? Aka all the manual tests together?

Keep in mind to always maintain a balance, by:
* making sure tests have a well defined purpose (description, expected results)
* the purpose is not too small, we don't want two many small tests, as the test setup overhead can increase a lot the time of the test runs.

test/check-timezone
```diff
@@ -0,0 +1,206 @@
...
+        i.open()
+        i.reach(i.steps.DATE_TIME)
+        dt.set_time_format("ampm")
+        dt.check_time_format("ampm")
+        dt.set_time_format("24")
+        dt.check_time_format("24")
+
+    def testTimezoneAndTimeformat(self):
+        """
+        Test full cycle of timezone, time format (24/ampm) and correct UTC backend saving.
```
KKoukiou, line 145: Keep the pattern for the docstrings always with 'Description.. \n Expected results:\n;'

src/components/timezone/DateAndTime.jsx
```diff
@@ -0,0 +1,458 @@
...
+                        .filter(Boolean);
+
+                const finalServers = ntps.length
+                    ? ntps
+                    : (confNtpServers.length ? confNtpServers : DEFAULT_NTP_SERVERS);
+
+                setNtpServers(finalServers);
+                setNtpServer(finalServers[0] || "ntp.fedoraproject.org");
+            } catch (err) {
+                console.error("Failed to fetch time/date/time source data:", err);
```
KKoukiou, line 109: We should not silently ignore errors here I think. If you dont handle it at all it would be cought by the global error handler if it fails, which I believe it better than ignoring.

src/components/timezone/DateAndTime.jsx
```diff
@@ -0,0 +1,458 @@
...
+
+    const handleAutoDateTime = useCallback(() => {
+        const newValue = !autoDateTime;
+        setAutoDateTime(newValue);
+        setNTPEnabled({ enabled: newValue });
+    }, [autoDateTime]);
+
+    // Section validity update
+    useEffect(() => {
+        setSectionValid(autoDateTime || datetime.isValid);
```
KKoukiou, line 124: Can you ensure that if the form has invalid data the next is disabled, and test it in the tests?

src/components/timezone/DateAndTime.jsx
```diff
@@ -0,0 +1,458 @@
...
+                <Stack hasGutter>
+                    <StackItem>
+                        <Flex alignItems={{ default: "alignItemsCenter" }}>
+                            <FlexItem>
+                                <Checkbox
+                                  id={`${SCREEN_ID}-auto-date-time`}
+                                  isChecked={autoDateTime}
+                                  onChange={handleAutoDateTime}
+                                  label={_("Automatically set date and time, using time server:")}
+                                  aria-label={_("Automatically set date and time")}
```
KKoukiou, line 196: If you're already using a visible label like here, adding aria-label is redundant and can confuse screen readers. 
Please remove.

garrett:
Apologies for not being around... I was on PTO for a couple weeks.

Thanks for implementing it based on what you had, even if it wasn't the state I handed it over before going on PTO. It's almost in the same state and _might_ only need a few changes. (Just going by the screenshot now.)

The obvious ones I can see from the screenshot:

1. AM/PM toggle should be moved as in the toggle mockups, for the reasons mentioned.

As far as the order: The page is a "Date and time" screen and that's why the heading is first. It's labeled as "Date and time" in the sidebar and this adds consistency.

But, yeah, timezone can impact the date and time. However, if someone sets the date and time and then adjusts the timezone, it doesn't _need_ to adjust the local time when the timezone changes. You're setting a local time first, then saying where you are (if it's not already correct based on rough location provided by network, if it exists).

So, yes, the timezone does adjust the local time if you represent the time internally as UTC and derive a local time from that and the timezone... but if you're setting the end-result local time first (which is what everyone actually sees anyway, on their watches, ovens, car dashboards, etc.) and doing the timezone second, then the internal UTC based time would change instead. Therefore, changing the timezone does not influence the actual local time that people interact with, but actually the internal UTC time instead. And the order then doesn't matter, as we're setting local time and timezone and deriving internal UTC time from that (instead of the other way around).

PR #862, opened 2025-06-06
components: storage: hide all scenarios when disk selection is not valid

KKoukiou:
Partially resolves: https://issues.redhat.com/browse/INSTALLER-4084

Before:

Screen Shot 2025-06-10 at 11 02 48

Screen Shot 2025-06-10 at 11 02 40

After:

Screen Shot 2025-06-10 at 11 01 56

Screen Shot 2025-06-10 at 11 02 06


rvykydal (approved):
Nice improvement, looks good to me.

(I was just hesitating about removing the availability setting from the scenario check. It felt to me as a kind of self-contained property that may be useful in general, independent of the higher level layout at this place, but maybe it is not.)

PR #863, opened 2025-06-06
components: storage: fix unselecting the only available disk

KKoukiou:
Previously, unselecting the only available disk and saving would reselect it automatically. That happened because, Arrays (even with the same content) are new references on every render. Convert the array to string to avoid unwanted re-renders.


rvykydal (approved):
Looks good to me. And good to know.

PR #870, opened 2025-06-11
test: avoid test failures because pool has been defined

KKoukiou:
Avoid errors with the following pattern:

ERROR    Error: --disk /var/tmp/*.qcow2,format=qcow2,target=vda: Could
not define storage pool: operation failed: pool 'tmp' already exists
with uuid *

subprocess.CalledProcessError: Command '['virt-xml', '-c', 'qemu:///session', 'fedora-rawhide-boot-127.0.0.2-2201', '--update', '--add-device', '--disk', '/var/tmp/*.qcow2,format=qcow2,target=vda']' returned non-zero exit status 1.

### DIFF
--- a/test/run
+++ b/test/run
@@ -98,6 +98,15 @@ fi
 # Allow wiki-reporting even if test suite partially failed
 set +e
 
+# Create tmp storage pool to avoid parallel test runs to conflict
+echo "<pool type='dir'>
+  <name>tmp</name>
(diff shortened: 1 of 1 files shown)


rvykydal (approved):
Looks good to me.
(I guess there is no risk of (in) using /tmp/strorage-pool.xml file in parallel).

PR #900, opened 2025-07-02
UX improvements to enable easier interaction from Open QA tests

KKoukiou:


garrett (approved):
Thanks! Looks good. :+1: 

(I checked the menu height and the password focusing when root is selected.)

PR #910, opened 2025-07-07
Only require Firefox on Fedora Workstation

KKoukiou:
This ensures Firefox is only required when installing on Fedora Workstation, while still avoiding the dependency on RHEL and other Fedora variants.

Fixes: INSTALLER-4189


M4rtinK (approved):
Looks good to me. :)

PR #922, opened 2025-07-17
components: users: autogenerate username from full name

KKoukiou:


adamkankovsky (approved):
I tested it locally and it seems fine to me.

PR #939, opened 2025-07-28
test: temporarily pretend that GTK anaconda path is missing

KKoukiou:
To fake removal of anaconda-gui till
https://github.com/rhinstaller/anaconda/pull/6530 reaches the gated rawhide ISO in bots.


M4rtinK (approved):

PR #948, opened 2025-08-06
tests: explicitly invoke mdadm --assemble to trigger known failure on rawhide

KKoukiou:
This triggers the mdadm --assemble failure early in the test when running with recent rawhide kernels where the operation is currently broken.

Doing so ensures that the RAID-related tests fail consistently and can be reliably matched through a naughty in bots, since there is no other stable pattern of failure detection at this point.


martinpitt (approved):
Nice, epharisto!

PR #949, opened 2025-08-06
tests: adjust to removed arguments from ssh_connection in bots

KKoukiou:
See bots commit:

commit b365eb61d03b85314df3e7241547ed7915207d68
Author: Allison Karlitskaya <allison.karlitskaya@redhat.com>
Date:   Wed Jul 30 10:26:05 2025 -0400

    machine: rework ssh argument handling a bit

### DIFF
--- a/test/webui_testvm.py
+++ b/test/webui_testvm.py
@@ -43,7 +43,7 @@ def cmd_cli():
             # print ssh command
             ssh_args = f"-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p {machine.ssh_port}"
             print(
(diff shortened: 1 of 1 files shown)


KKoukiou:
@allisonkarlitskaya may I ask for a review on this?

allisonkarlitskaya (commented):
This is fine of course but there are likely to be future problems.

The `_get_ssh_options()` function is going to be the "one true way" to get the options for ssh.  We could make that public so you can call that instead.  Alternatively I think there's some diagnose() function or similar that can print the information.  All of the other attributes are privateish/semi-deprecated at this point and very likely to disappear soon...

PR #950, opened 2025-08-06
components: cockpit-storage-integration: fix warning banner not taking full width

KKoukiou:
Possibly when we ported the modals to Patternfly 6 this broke. Fix the button alignment and missing background by using a custom 'Close' button for the modal, to have full control over it.

Resolves: rhbz#2386742


garrett (approved):
Thanks! This looks a lot better!

PR #957, opened 2025-08-12
tests: ignore timezone in review screen as this is autodetected

KKoukiou:
And test runners do not always exist in the same data center.

Example failure: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-6549-a888a48b-20250812-071735-fedora-rawhide-boot-anaconda-pr-6549-rhinstaller-anaconda-webui/TestStorageMountPoints-testMultipleDisks-review-multiple-disks-delta.png


adamkankovsky (approved):
I didn't even think that could happen. Thanks

PR #958, opened 2025-08-12
Size check mpa review

rvykydal:


rvykydal:
@garrett may I ask for a review?
One thing that I noticed already in the initial version is mentioning 'on assigned mount points' in the message. Either we should say something neutral, generic, like 'Not enough free space for the system' or enumerate the checked mount points?

The cockpit storage tests are failing, need to look at it, but I think we can start with the UI review.

rvykydal:
> OK, it's excellent that we're checking for this! Thanks for thinking about it and implementing it!
> 
> I do think it can be improved in a few ways:
> 
>     1. Have warnings/errors **wherever people are setting up partitions**.
>        
>        * This would be everywhere partitions are worked with. Mount point assignment, Cockpit Storage, and even when selecting one of the scenarios... if there isn't enough space.
> 

Agreed that it would be better. As mentioned in the Note in the description, the best we could get to technically was  https://github.com/rhinstaller/anaconda-webui/pull/882 and it was not accepted for technical/architectural reasons.

>     2. Show the problems **when the partitioning is being checked**.
>        
>        * (Checking partitions inline while they're being edited. And during the after-configuration check as well.)
> 

We are not checking individual partitions, the API (backend) is only able to checking the whole partitioning (which is created when mount points are collected and the partitioning is created. Checking when individual partition is edited is not easily possible at the moment.

>     3. On this specific review screen, it would be better to have it **closer to the problem**.
>        
>        * Right now, it's at the top, but this is something that could go in the storage area. Not just that, it could actually be contextual where it is next to the actual partitions that have problems.
>        * This screenshot makes it look like `/boot` is a problem too, but it's really just an issue with `/`. _(As an example: The work laptop I'm typing this on has < 1 G `/boot/`: `df -h /boot` shows: `/dev/nvme0n1p2`  Size: `974M`  Used: `681M`  Avail: `226M`  `76%`.)_
>        * We could still have an alert at the top (as this is a critical issue), but if so, then it should be generic and then we would still need it to be contextual as to where the problem is.
> 

I think this could be really improved in scope of this issue / PR.
We have a list of partitions that are checked for the space ('/', '/usr' currently).

>     4. If we know this _will_ be a problem, then it shouldn't be a warning, **it should be an error**.
>        
>        * The error should be resolved (by the person using Anaconda) before proceeding.
>        * We shouldn't let people go forward until the error is fixed, as it will erase their current system (which might work) and replace it with a failed installation that will not work.
>        * There should be a clear way to say how to fix it, and provide a way to fix it. This is why catching it earlier (first two points) is important. But here on the review screen, we would need to have a way to go back to fix it. A link back to the page where it was set might work. Plain text stating that someone needs to go back to change it would also be OK, even if it doesn't provide the solution, it would at least be a hint for what to do.

The only possible improvement I see in scope of this issue is adding a hint.
Regarding the error vs warning and preventing proceeding - the reason is in the description. We don't want prevent proceeding. That is also why I used Warning and not Error alert.

rvykydal:
The PR is not critical for F43. An important part issue was (if I am getting it right) that even developers don't get the reason of failed installation easily. So we could just go with logging (https://github.com/rhinstaller/anaconda-webui/pull/844) for F43 and try to do this better for F44.
I am starting to think we don't have enough time for valuable solution in F43 now.

garrett:
A "best effort for now" mockup:

Wizard, review, table

Again, ideally, we'd catch this in the same place where people can make the changes, but this at least addresses the issues in the review screen and is better than having someone wipe out their (possibly working) installation with a failed install.

rvykydal:
@garrett looks definitely better. Have some additional questions:

1) Where the 'change partitioning default' link points to? I think `Installation method` screen.
2) What if /usr is a saparate partition, i.e. we have `/` and `/usr` in the list. I am not able to come up with something good local. Maybe we can just skip it in this (not so frequent, otoh the warning would appear usually in less usual cases - like cockpit storage / custom partitioning) case?
3) Should I really use Error (vs Warning) given we allow to proceed with the installation?
4) Should we make proceeding with the installation more difficult or different? (another label on the button, added checkbox...)

garrett:
1. The link would point to the place where you configure the storage. I guess at least in this case, it would point to mount point assignment (according to the button, which should have different wording... but that's another thing to improve outside of this PR).

2. If we cannot determine which partitions are causing the problem, at least the message above is more vague where it doesn't seem like it's an specific issue with all partitions under that size (as the first screenshot implies).

3. My question is: Should we really allow someone to proceed with the installation? At this point, we _know_ it'll fail and it _will_ cause data loss. Therefore, yes. It should be an error and you shouldn't be able to proceed.

4. Probably not. This is hopefully a corner case? Is there any reason at all why someone would have an install so small in the year of 2025? For a working Linux system (or, really: any OS), you need a little bit of extra space anyway, so even if the estimate is a little small, then it will still have problems quite quickly after any bit of use.

rvykydal:
> I'm concerned that we're making it too easy to have people attempt to install in a situation where we _know_ it will fail, and they will wind up having a broken system... and worse yet, losing any data they might've had on the system prior (if the disk is not already blank).
> 

For desktop use case I agree with error / preventing the installation.

rvykydal:
> Can't we do a size comparison when constructing the table that shows the partitions and compare the size that is set for each partition with the minimum, and show the errors next to any of the ones that do not meet the minimum, provided they're either within a certain set? I'd imagine anything that's from actual storage (not tmpfs, proc, sys, etc.) mapped in root aside from boot, home, mnt, srv, etc is probably subject to this.

Who would provide and maintain the table - set the minimum?
Having such data part of project makes its maintenance significantly more difficult. 

How would be such minimums determined, for example for the /usr ? The size of the whole payload that ends up in /usr is dynamic.
We have static requirements that can be used for OS variants - for example [for ovirt](https://github.com/rhinstaller/anaconda/blob/fb3114315e6c43b9f7adc1902fe371eb2768c63a/data/profile.d/ovirt.conf#L29). But these are using a different mechanism (static) and have a different purpose.

> 
> That is, if we consider this listing of all top-level subdirectories in /:
> 
> bin boot dev etc home lib lib64 media mnt opt ostree proc root run sbin srv sys sysroot tmp usr var
> 
> That means if any of these are mounted from another place and are too small (whatever the actual limit is, regardless of what we might think it is), the system will fail: /bin, /etc, /lib, /lib64, /ostree /sbin, /usr, /var
> 
> Of those, it's going to really be /usr and /var... and / itself. So that's really 3 we need to care about here. The others are very likely not going to be mounted from some too-small partition.

Yes, and that is the problem we are trying to address in this PR.
For these there are no static limits, we are using the payload API to get the size required for the actually selected payload. We need to keep in mind broader context then Desktop using image payload (rsync) with preselected package set.

> 
> (/boot and /boot/efi do also matter too, but they have different space constraints than the rest of the system.)

Handled by backend static constraints: https://github.com/rhinstaller/anaconda/blob/main/pyanaconda/modules/storage/platform.py#L228 which currently result in warning during "partitioning configuration" (API term) (for example leaving MPA screen).

rvykydal:
> Another issue: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-958-a8ee9ea3-20250818-115937-fedora-rawhide-boot-cockpit/log.html#9 (strangely, the pixel test does not fail locally for me although the layout changes the same way)
> 
> pixel: TestStorageCockpitIntegration-testLVM-cockpit-storage-integration-check-storage-dialog-LVM-pixels
> 
> reference: TestStorageCockpitIntegration-testLVM-cockpit-storage-integration-check-storage-dialog-LVM-reference

Fixed in the last push by using the helper text only on review screen (where the Alert/error is shown).

src/components/review/StorageReview.jsx
```diff
@@ -52,10 +54,30 @@ import "./StorageReview.scss";
...
+                _("Not enough free space on assigned mount points. Required: $0 Available: $1."),
+                cockpit.format_bytes(requiredSize),
+                cockpit.format_bytes(freeSpace)
+            );
+            setStepNotification({
+                message: msg,
+                step: SCREEN_ID,
+                variant: "warning"
+            });
+        } else if (setStepNotification) {
```
KKoukiou: I would prefer the setStepNotification check in the start if the useEffect:

if (setStepNotification === undefined) { return };

src/components/review/StorageReview.jsx
```diff
@@ -54,30 +57,10 @@ import "./StorageReview.scss";
 
 const _ = cockpit.gettext;
 const idPrefix = "storage-review";
-const SCREEN_ID = "anaconda-screen-review";
```
KKoukiou: Removed by mistake :) ?

KKoukiou (approved):
Some minor comments

src/components/review/ReviewConfiguration.jsx
```diff
@@ -87,18 +94,64 @@ const AccountsDescription = () => {
...
+            const actionLinks = (
+                <Button
+                  id={`${SCREEN_ID}-change-partition-layout`}
+                  variant="link"
+                  isInline
+                  onClick={() => {
+                      // Reset form state like the wizard does during navigation
+                      setIsFormValid(false);
+                      setIsFormDisabled(true);
+                      cockpit.location.go([fixupPage]);
```
KKoukiou, line 134: I am not sure why it's not working but I tried this and this works:

```diff
diff --git a/src/components/review/ReviewConfiguration.jsx b/src/components/review/ReviewConfiguration.jsx
index 8d3276e2..9de89b07 100644
--- a/src/components/review/ReviewConfiguration.jsx
+++ b/src/components/review/ReviewConfiguration.jsx
@@ -23,6 +23,7 @@ import {
     DescriptionList,
     Flex, FlexItem,
     Stack,
+    useWizardContext,
     useWizardFooter,
 } from "@patternfly/react-core";
 
@@ -107,6 +108,7 @@ export const ReviewConfiguration = ({ setIsFormValid, setStepNotification }) =>
     const freeSpace = useFreeSystemMountPointsSpace();
     const requiredSize = useRequiredSize();
     const { setIsFormDisabled } = useContext(FooterContext);
+    const { goToStepById } = useWizardContext();
 
     useEffect(() => {
         const step = SCREEN_ID;
@@ -131,7 +133,7 @@ export const ReviewConfiguration = ({ setIsFormValid, setStepNotification }) =>
                       // Reset form state like the wizard does during navigation
                       setIsFormValid(false);
                       setIsFormDisabled(true);
-                      cockpit.location.go([fixupPage]);
+                      goToStepById(fixupPage);
                   }}
                 >
                     {_("Change partition layout")}
```

rvykydal:
The test failed on F43 https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-958-ecae5ae2-20250821-110101-fedora-43-boot-storage/log.html
`wait_js_cond(ph_in_text("#helper-disk-vda[data-path='/'] .pf-v6-c-helper-text__item.pf-m-error","Needs at least 2.60 GB")): Error: actual text: Needs at least 2.61 GB
`
Need to make the test less rigid.

rvykydal:
Tests are passing now.
I had to keep the
`cockpit.location.go([fixupPage]);` before the `goToStepById` call
so that `windows.location.hash` gets updated correctly (the `Installer.reach()` failed in the test because there was still `anaconda-storage-review` hash)

I am also working on a test case for Cockpit Storage partitioning (the current is for MPA) but have some issues so will add it as a separate follow-up.

rvykydal:
> I am also working on a test case for Cockpit Storage partitioning (the current is for MPA) but have some issues so will add it as a separate follow-up (it works manually but not automated, perhaps some sync issue there).

The test added in the commit in https://github.com/rhinstaller/anaconda-webui/pull/974 fails for me [here](https://github.com/rvykydal/anaconda-webui/blob/efa8a150712b177674e30009d9a7579895ea726f/test/check-storage-cockpit#L337), reporting 0B of available space. It seems in general that the [hook](https://github.com/rvykydal/anaconda-webui/blob/efa8a150712b177674e30009d9a7579895ea726f/src/hooks/Storage.jsx#L87) does not update the value correctly with cockpit storage mount points directly used. Failed for me also when I tried to replace the `plannedDeviceTree` hook dependency with `appliedPartitioning`. There must be some sync issue (the device tree not being ready to read the value?). The case worked for me manually if I went through slowly. Without /boot partition even if I went through normally. Needs more investigation but the feature seems to be broken for cockpit storage scenario.

rvykydal:
Added the tests from https://github.com/rhinstaller/anaconda-webui/pull/974 to this PR.

For `ext4` the [tests](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-958-0e0c715f-20250826-070232-fedora-43-boot-cockpit/log.html) (`TestStorageCockpitIntegration.testPayloadSizeCheckExt4`) works except for unrelated failure which will have to investigate.

For `xfs` the test (`TestStorageCockpitIntegration.testPayloadSizeCheckXFS`) fails because the free space on the partition is detected as 0. The root cause is blivet not being able to detect minimum size of the device:
Log message: `WARNING:blivet:Failed to obtain minimum size for device /dev/vda3:`
The code: https://github.com/storaged-project/blivet/blob/ade42281451e2b0eccb5527c7e603b6d07a20797/blivet/formats/fs.py#L387

The code path affects devices with existing format: https://github.com/rhinstaller/anaconda/blob/276c9ce614aacbfbea36a7ace64757a67dcdd41a/pyanaconda/modules/storage/devicetree/model.py#L178 -> https://github.com/storaged-project/blivet/blob/ade42281451e2b0eccb5527c7e603b6d07a20797/blivet/formats/fs.py#L414 -> https://github.com/storaged-project/blivet/blob/ade42281451e2b0eccb5527c7e603b6d07a20797/blivet/formats/__init__.py#L410 where `_min_instance_size` is the same as the current_size due to the beforementioned failure: https://github.com/storaged-project/blivet/blob/ade42281451e2b0eccb5527c7e603b6d07a20797/blivet/formats/fs.py#L373

So the affected cases (existing format) in webui are:

- this case - the size check for a xfs partition created in cockpit storage scenario (Use configured storage) blocks installation (the device is formatted by cockpit storage so the manual partitioning requests has `reformat=False`, ie existing format)
```
Aug 25 14:19:42.031533 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com org.fedoraproject.Anaconda.Modules.Storage[6379]: DEBUG:anaconda.modules.storage.partitioning.manual.manual_module:Requests are set to '[MountPointRequest(device_spec='vda1', format_options='', format_type='biosboot', ks_spec='', mount_options='', mount_point='', reformat=False), MountPointRequest(device_spec='vda2', format_options='', format_type='ext4', ks_spec='', mount_options='', mount_point='/boot', reformat=False), MountPointRequest(device_spec='vda3', format_options='', format_type='xfs', ks_spec='', mount_options='', mount_point='/', reformat=False)]'.
```
- shrinking existing xfs partition (there is no test for it currently)
- home reuse with existing '/' on xfs - the check would block installation (not sure if it is supported)

rvykydal:
@vojtechtrefny can you please look at previous comment?
Especially the fact that `_minsize_class` is not implemented for `XFS` - is it expected? Can it be implemented?

rvykydal:
> For `ext4` the [tests](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-958-0e0c715f-20250826-070232-fedora-43-boot-cockpit/log.html) (`TestStorageCockpitIntegration.testPayloadSizeCheckExt4`) works except for unrelated failure which needs to be investigated.

Apparently there is some sync issue (race condition, hit also with `xfs` locally) when going forward through Accounts screen after updating the storage layout in Cockpit Storage. Actually for the feature it is not critical to test the size check error disappears (check passes) after changing the layout. The size check passing is tested by common / basic tests. The size check passing after fixing the layout (disappearing from the review screen) is tested in mount point assignment test.

Test of update of the layout in cockpit storage should probably have its own independent test not blocking this feature.

KKoukiou (approved):

PR #969, opened 2025-08-20
infra: add matrix strategy to test-compose workflow

KKoukiou:
- Add matrix to test against both rawhide and Fedora 43 when no specific compose_id is provided
- Maintain manual runs with specific compose_id


abadger (approved):
Approved

PR #983, opened 2025-09-01
users: disallow proceeding from users page when neither root or user is configured

KKoukiou:
The users need to create an admin user account *or* set a root password, but you they do not have to do both.

Relevant to: rhbz#2379957
Resolves: rhbz#2392205
Resolves: rhbz#2392203


adamkankovsky (approved):
Looks great for me, This is definitely useful because it caused problems on some spins as well.

PR #987, opened 2025-09-05
tests: fedora-41 image was dropped, let's move to using fedora-42 instead

KKoukiou:


martinpitt (approved):
Thank you! This is both urgent and also obvious, so let's land this ASAP.

PR #991, opened 2025-09-09
fix(timezone): correctly parse timezones with multiple slashes

snowyyd:
This should resolve the `RangeError` that is thrown during the installation of Fedora 43 when the time zone contains multiple slashes.

For example, a `RangeError` is thrown when the timezone is: `America/Argentina/Buenos_Aires`.

Example:

```ts
const timezone = 'America/Argentina/Buenos_Aires';

const [reg, cty] = timezone.split('/');
console.log(reg, cty); // output: America Argentina

const [reg, ...cty] = timezone.split('/');
console.log(reg, cty.join('/')); // output: America Argentina/Buenos_Aires
```

### DIFF
--- a/src/components/datetime/Timezone.jsx
+++ b/src/components/datetime/Timezone.jsx
@@ -69,9 +69,9 @@ export const TimezoneSection = ({ locale, setSectionValid, setTimezoneLabel }) =
 
     useEffect(() => {
         if (timezone && typeof timezone === "string" && timezone.includes("/")) {
-            const [reg, cty] = timezone.split("/");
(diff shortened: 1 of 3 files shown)


KKoukiou (commented):
thanks for the fix, I added a test.

KKoukiou:
Added a second fix on the timezone screen which I noticed while testing this.

KKoukiou:
Let's land the 3 commits in seperate PR: https://github.com/rhinstaller/anaconda-webui/pull/993 and then I will merge this.

PR #999, opened 2025-09-12
tests: remove compose from the ResTuple

KKoukiou:
Setting `compose='20250910.n.0'` is wrong. It's not breaking anything because wikitcms [1] uses cid if present and ignores any release/milestone/compose values. However, release/milestone/compose and cid are alternative ways to try and identify the thing we're reporting for.

[1] https://pagure.io/fedora-qa/python-wikitcms/blob/main/f/src/wikitcms/wiki.py


M4rtinK (approved):
Looks good as far as my limitted understanding of this allows. :)

PR #1023, opened 2025-10-01
tests: introduce new test for DNF payload consumed from bots repository

KKoukiou:


M4rtinK:
As this changes how the UI looks like (presume it adds a new screen to the wizard)  I think it would be good to add some screenshots. :)

KKoukiou:
> As this changes how the UI looks like (presume it adds a new screen to the wizard) I think it would be good to add some screenshots. :)

Added pixel tests, let me link these here.
https://github.com/rhinstaller/pixel-test-reference/compare/74084465421a2205a831375595ee4b46ff0f4d6f..c5100e57533612fd87323b451c88bb89ad68d3c6

bruno-fs (approved):

PR #1030, opened 2025-10-07
tests: /boot is now by default 2GB

KKoukiou:
Related: https://github.com/rhinstaller/anaconda/pull/6678


rvykydal (approved):
Looks good to me.
We should add a test for the now failing case that would reflect its handling (mess).

PR #1151, opened 2026-01-27
Change default boot mode to UEFI for tests

KKoukiou:
Update all references to use inverted logic.
Also remove @run_boot('efi') decorators from test files as EFI is now default.

**To be landed together with: https://github.com/cockpit-project/bots/pull/9050**


M4rtinK (approved):
Looks good to me as far as I can tell, thanks! :)

PR #1186, opened 2026-02-26
tests: add network test for 'Wired Connection' assumptions in Cockpit

rvykydal:
Related: COCKPIT-1750


KKoukiou (commented):
Does this check really belong to WebUI? Is there anything WebUI does to affect this? I don't see any UI visible data either.

PR #1204, opened 2026-03-04
storage: Add tooltips to Reclaim dialog partition action icons

KKoukiou:
- Attach resize tooltip via triggerRef instead of wrapping the button, so the Popover still opens on click (wrapping with Tooltip breaks the Popover trigger).
- Use controlled Popover (isVisible + shouldOpen/shouldClose) so the shrink popover does not open when the button is aria-disabled. PatternFly Popover still fires onTriggerClick for aria-disabled triggers, so we only call show() from shouldOpen when !isAriaDisabled.

Resolves: rhbz#2403027


adamkankovsky (approved):
Looks great to me. Thanks

PR #1207, opened 2026-03-06
log: serialize objects with JSON.stringify so errors show content in logs

KKoukiou:


rvykydal (approved):

bruno-fs (approved):

KKoukiou (commented):
Error objects, which are printed in case of exceptions are not enumerable to JSON.stringify does not print them. 
I needed to create a overrride method for that.

PR #1225, opened 2026-03-23
Network configuration in wizard page

rvykydal:
Draft for https://redhat.atlassian.net/browse/INSTALLER-4650

Requires https://github.com/rhinstaller/anaconda/pull/6989

If we follow on this path:

- [x] fix tests
- [x] add tests
- [x] ~remove network configuration from header menu. (Maybe it would be good to keep it there as an addon to a new simple connectivity UI in Wizard)~
- [x] rebase upon ks support

Notes:

- the patchset disables navigation elements also for iframe modal dialogs but not all of them pass the event to anaconda: https://github.com/cockpit-project/cockpit/pull/22963 (in the video: the "Keep Connection" dialog does, the MTU setting dialog does not.)
- The PR / solution does not seem to be very viable as it seems we will move towards integration of a component provided by Cockpit. (UPDATE: or maybe rather keep using iframe integration and parametrize the screen for Anaconda). Although I am sceptical about minimal Wizard page being sustainable - I can see requests for at least adding virtual devices configuration appearing quickly - it could make sense to have minimal network page (mainly for Wireless configuration, maybe even anaconda-webui specific code if it does not fit to Cockpit, or special anaconda-wizard mode of cockpit module) and keep the access to the Cockpit module available (from the wizard page or even the page header menu as it is currently).
- The part disabling the navigation can be useful also for different (non-iframe) approaches. UPDATE: it was moved to a separate PR: https://github.com/rhinstaller/anaconda-webui/pull/1231

Video:

NOTE: the disabling of navigation on sidebar on the video is now moved to a separate PR: https://github.com/rhinstaller/anaconda-webui/pull/1231

https://github.com/user-attachments/assets/4b13cc59-5017-4205-8116-cf8433c52127

### DIFF
--- a/src/components/network/NetworkConfiguration.jsx
+++ b/src/components/network/NetworkConfiguration.jsx
@@ -40,6 +40,21 @@ export const NetworkConfiguration = ({
             iframe.contentWindow.addEventListener("error", exception => {
                 onCritFail({ context: _("Network plugin failed") })({ message: exception.error.message, stack: exception.error.stack });
             });
+
+            // Hide elements not needed in the installer context
+            const hideSelectors = ["#networking-graphs", ".cockpit-log-panel"];
+            const iframeDoc = iframe.contentDocument;
+            const observer = new MutationObserver(() => {
+                hideSelectors.forEach(sel => {
+                    const el = iframeDoc.querySelector(sel);
+                    if (el && el.style.display !== "none") {
+                        el.style.display = "none";
+                    }
+                });
+            });
+            observer.observe(iframeDoc.body, { childList: true, subtree: true });
+
+            return () => observer.disconnect();
         }
     }, [isIframeMounted, onCritFail]);
 
--- a/src/components/network/NetworkConfiguration.jsx
+++ b/src/components/network/NetworkConfiguration.jsx
@@ -19,7 +19,7 @@ const _ = cockpit.gettext;
 export const NetworkConfiguration = ({
     onCritFail,
 }) => {
-    const { setIsFormValid } = useContext(PageContext) ?? {};
+    const { setIsFormDisabled, setIsFormValid } = useContext(PageContext) ?? {};
     const [isIframeMounted, setIsIframeMounted] = useState(false);
     const { hasActiveCheckpoint } = useNetworkStatus();
     const backdropClass = useMaybeBackdrop();
@@ -28,7 +28,8 @@ export const NetworkConfiguration = ({
 
     useEffect(() => {
         setIsFormValid(!hasActiveCheckpoint);
-    }, [hasActiveCheckpoint, setIsFormValid]);
+        setIsFormDisabled(hasActiveCheckpoint);
+    }, [hasActiveCheckpoint, setIsFormDisabled, setIsFormValid]);
 
     useEffect(() => {
         if (isIframeMounted) {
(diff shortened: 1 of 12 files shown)


rvykydal:
Rebased on top of https://github.com/rhinstaller/anaconda-webui/commit/3f227d0a517082889bac8dac617007ce782ae004

src/components/network/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+const _ = cockpit.gettext;
+
+export class Page {
+    _description = "Configure network connections for the system.";
+
+    constructor ({ isBootIso }) {
+        this.component = NetworkConfiguration;
+        this.id = "anaconda-screen-network";
+        this.isHidden = !isBootIso;
```
KKoukiou, line 18: I guess this is copied from elsewhere, but the correct way to do it is to update https://github.com/rhinstaller/anaconda/blob/main/data/profile.d/fedora-workstation.conf#L28 and it will be automatically picked up.

src/components/network/NetworkConfiguration.jsx
```diff
@@ -0,0 +1,71 @@
...
+ */
+import cockpit from "cockpit";
+
+import React, { useContext, useEffect, useState } from "react";
+
+import { PageContext } from "../../contexts/Common.jsx";
+
+import { useMaybeBackdrop } from "../../hooks/CockpitIntegration.jsx";
+
+import { useNetworkStatus } from "./useNetworkStatus.js";
```
KKoukiou, line 13: No hard objection, but we do place hooks usually in src/hooks/[Network.jsx]

src/components/network/useNetworkStatus.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+import { useEffect, useState } from "react";
+
+// Hook to track checkpoint status from the Cockpit networkmanager iframe
+export const useNetworkStatus = () => {
+    const [hasActiveCheckpoint, setHasActiveCheckpoint] = useState(false);
+
+    useEffect(() => {
+        const checkpointState = window.sessionStorage.getItem("cockpit_has_checkpoint");
+        setHasActiveCheckpoint(checkpointState === "true");
```
KKoukiou, line 13: We have a helper parseAnacondaConfBool in src/helpers/utils.js which convert such values to bool.

src/components/network/NetworkConfiguration.jsx
```diff
@@ -40,6 +40,21 @@ export const NetworkConfiguration = ({
             iframe.contentWindow.addEventListener("error", exception => {
                 onCritFail({ context: _("Network plugin failed") })({ message: exception.error.message, stack: exception.error.stack });
             });
+
```
KKoukiou, line 43: While this in an interesting hack, it's not consistent to how we hide stuff in the cockpit-storage. 

For that we have anaconda.css in the cockpit-storage plugin and apply it.

KKoukiou (approved):
Some comments, nothing blocking. 
This will conflict with the KS work. If we merge this first, which is fine, I will update the KS support PR.

src/components/network/NetworkConfiguration.jsx
rvykydal: You mean the code below. Seems to be handled by https://github.com/rhinstaller/anaconda-webui/pull/1269 and the related Cockpit patch now, thank you.
I'd revert the "network: show only Interfaces part of Cockpit screen" from this set when the cockpit change is in.

KKoukiou:
There is some conflict in the tests. I will rebase for you.

PR #1231, opened 2026-03-26
Really disable navigation on wizard side navigation bar

rvykydal:
We should disable the navigation on wizard sidebar for real when isFormDisabled is true. Currently it is still possible to click away from the screen on the sidebar.

This is problematic for example when checkpoint is in progress in network configuration, where destroying the iframe can lead to broken connectivity.

Related: INSTALLER-4650

The problem: see on the video below that it is possible to move away from the screen during a checkpoint in progress via sidebar navigation:

The desired state is on [the video in a PR](https://github.com/rhinstaller/anaconda-webui/pull/1225#issue-4121956073) from which I split out this one.

https://github.com/user-attachments/assets/0b6f016f-ee72-42b4-908d-bd9381cfc493

### DIFF
--- a/test/helpers/installer.py
+++ b/test/helpers/installer.py
@@ -178,6 +178,7 @@ class Installer():
 
     def click_step_on_sidebar(self, step=None):
         step = step or self.get_current_page()
+        self.browser.wait_visible(f"#{step}:not([disabled])")
         self.browser.click(f"#{step}")
 
     @log_step()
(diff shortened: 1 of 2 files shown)


test/helpers/installer.py
```diff
@@ -178,6 +178,7 @@ def open(self, step=None):
 
     def click_step_on_sidebar(self, step=None):
         step = step or self.get_current_page()
+        self.browser.wait_visible(f"#{step}:not([disabled])")
```
KKoukiou, line 181: This is not needed but it does not hurt. b.click (the operation after this) and all operations wait for elements to not be disabled before they perform actions.

KKoukiou (approved):

test/helpers/installer.py
rvykydal: Unfortunately I don't remember why I added this. I guess some tests were not passing - but given the click should wait as well that would be strange and maybe indicating some other problem.
Maybe I will try to remove it and run the tests?

KKoukiou: I am  ok to leave it here.

PR #1242, opened 2026-04-07
infra: Webui sizzle remove

adamkankovsky:


KKoukiou (commented):
I think the last commit might not be needed? 


I see cockpit supports it's own contains: 

test/common/test-functions.js:            // Best effort support `:contains()`

adamkankovsky:
I don't think the declining test scores have anything to do with the change.

KKoukiou (commented):
The failing tests look relevant. Please rebase and see actual results.

In any case, the test suite was stabilized today, if you see errors after rebasing that must be a problem here.

PR #1247, opened 2026-04-09
tests: workaround wrongly detected 'use-configured-storage-kickstart' scenario

KKoukiou:
The new 'Storage configured via kickstart' scenario, automatically gets preset when the is one created partitioning when loading the app. This is safe guess for real world scenarios and also mimics the GTK UI behavior.

The non-destructive tests, as they can't clean up created partitioning sometime hit this issue causing flakes by wrong scenario selection.

To fix this, make sure that there is never a single partitioning when starting the nondestructive tests, by not creating it in the storage cleanup.


rvykydal (approved):
Looks good to me.

PR #1262, opened 2026-04-20
tests: remove redundant root btrfs volume mount in cockpit tests

KKoukiou:


mvollmer (approved):
Thanks! This should make https://github.com/cockpit-project/cockpit/pull/23149 green.

PR #1265, opened 2026-04-21
Drop Sizzle and align tests with Cockpit testlib without Sizzle

KKoukiou:
Remove the unused Sizzle devDependency and refresh the node_modules cache.

Cockpit test-functions only supports a single :contains() per selector when Sizzle is absent.

- reclaim: add data-device attribute to table rows and rewrite StorageReclaimDialog helpers to use CSS selectors instead of JS IIFEs
- review: assert disk table rows via wait_js_cond over #storage-review-table-<disk> tbody tr text
- network: call NetworkCase helpers through a SimpleNamespace(browser=...) so behaviour matches netlib without duplicating [data-label] selectors

### DIFF
--- a/src/components/review/StorageReview.jsx
+++ b/src/components/review/StorageReview.jsx
@@ -130,18 +130,24 @@ const DeviceRow = ({ disk, isReviewScreen }) => {
                 )
                 : ""
         );
-
+        const encryptedText = hasEncryptedAncestor(devices, device) ? (isReformattedMountPoint ? _("encrypt") : _("encrypted")) : "";
+        const deviceText = cockpit.format("$0$1", parents.join(", "), showMaybeType());
         return (
             {
                 columns: [
-                    { title: cockpit.format("$0$1", parents.join(", "), showMaybeType()), width: 17 },
+                    { title: deviceText, width: 17 },
                     { title: size, width: 15 },
                     { title: action, width: 17 },
-                    { title: hasEncryptedAncestor(devices, device) ? (isReformattedMountPoint ? _("encrypt") : _("encrypted")) : "", width: 17 },
+                    { title: encryptedText, width: 17 },
                     { title: mount, width: 17 },
                     ...(isReviewScreen ? [{ title: helperText, width: 17 }] : []),
                 ],
-                props: { key: device },
+                props: {
+                    "data-device": deviceText,
+                    "data-mount": mount,
+                    ...(encryptedText ? { "data-encrypted": encryptedText } : {}),
+                    key: device
+                },
             }
         );
     };
@@ -177,6 +183,8 @@ const DeviceRow = ({ disk, isReviewScreen }) => {
                     ...(isReviewScreen ? [{ title: "" }] : []),
                 ],
                 props: {
+                    "data-action": actionDescriptionText,
+                    "data-device": device,
                     key: device + actionType,
                 },
             }
--- a/test/check-storage-reclaim
+++ b/test/check-storage-reclaim
@@ -75,7 +75,7 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
 
         # Test cancelled reclaim is not taking effect
         s.reclaim_remove_device("vda2")
-        s.reclaim_check_action_present("vda2", "delete", True)
+        s.reclaim_check_action_present("vda2", "delete")
         s.reclaim_modal_cancel()
 
         # Continue without reclaiming space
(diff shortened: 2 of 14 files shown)


test/helpers/review.py
```diff
@@ -55,23 +55,33 @@ def check_disk_row(
...
-            f"td:contains('{action}') + "
-            f"td:contains('{encrypt_text}') + "
-            f"td:contains('{mount_point}')"
-        )
+
+        # Action rows (delete/resize) have data-action and no data-mount;
+        # mount rows have data-mount and no data-action.
+        # Detect action rows: explicit non-default action, no mount_point, no reformat.
+        is_action_row = action and action not in ("mount", "biosboot") and not mount_point and not reformat
+        if is_action_row:
```
KKoukiou, line 71: I think you want to check all passed, aka it's possible you want to check both mount point and action. in this case you dont want to use elif here.

test/helpers/review.py
```diff
@@ -55,23 +55,38 @@ def check_disk_row(
...
+        # mount rows use data-device and/or data-mount; delete/resize rows use data-action.
+        # A real row never has both data-action and data-mount; conflicting fragments match nothing.
+        is_action_row = action and action not in ("mount", "biosboot") and not reformat
+        parts = [f"{table} tr"]
+        if parent:
+            parts.append(f'[data-device="{parent}"]')
+        if mount_point:
+            parts.append(f'[data-mount="{mount_point}"]')
+        if is_action_row:
+            parts.append("[data-action]")
```
KKoukiou: You are missing here the action type.

test/helpers/review.py
```diff
@@ -55,23 +55,38 @@ def check_disk_row(
...
+        elif parent and not mount_point:
+            # e.g. biosboot: device row with empty mount path — still has a data-mount attribute
+            parts.append("[data-mount]")
```
KKoukiou: You are missing here the mount point.

test/helpers/review.py
```diff
@@ -55,23 +55,33 @@ def check_disk_row(
...
+        elif mount_point:
+            row = f'{table} tr[data-mount="{mount_point}"]'
+        else:
+            row = f'{table} tr[data-device="{parent}"][data-mount]'
+
+        self.browser.wait_visible(row)
+
+        for text in (parent, str(size) if size != "" else "", action_text, encrypt_text, mount_point):
+            if text is not None and str(text) != "":
+                self.browser.wait_in_text(row, str(text))
```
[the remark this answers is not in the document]
KKoukiou: @adamkankovsky valid I guess.

test/helpers/storage.py
```diff
@@ -516,49 +516,57 @@ def set_scenario(self, scenario):
...
-    def reclaim_check_device_row(self, location, name=None, deviceType=None, space=None, locked=False):
-        self.browser.wait_visible(
-            "#reclaim-space-modal-table "
-            f"td[data-label=Location]:contains({location}) + " +
-            (f"td[data-label=Name]:contains({name}) + " if deviceType != "disk" else "") +
-            f"td[data-label=Type]:contains({deviceType}) + "
-            f"td[data-label=Space]:contains({space})"
-        )
+    def _reclaim_row_selector(self, device):
+        """CSS selector for the table row identified by data-device attribute."""
```
KKoukiou, line 525: The method is a bit self explanatory, no need for docstring imo.

test/helpers/storage.py
```diff
@@ -516,49 +516,57 @@ def set_scenario(self, scenario):
...
+        if deviceType is not None:
+            self.browser.wait_in_text(f'{row} td[data-label="Type"]', deviceType)
+        if space is not None:
+            self.browser.wait_in_text(f'{row} td[data-label="Space"]', space)
         if locked:
-            self.browser.wait_visible(
-                f"#reclaim-space-modal-table tr:contains({name}) "
-                f"td[data-label=Type] .reclaim-space-modal-device-locked"
-            )
+            self.browser.wait_visible(f'{row} .reclaim-space-modal-device-locked')
```
KKoukiou, line 544: For consistency with the above i would keep using "{row} td[data-label=Type]  .reclaim-space-modal-device-locked"

test/helpers/storage.py
```diff
@@ -516,49 +516,57 @@ def set_scenario(self, scenario):
...
-
-        if present:
-            self.browser.wait_visible(selector)
+        sel = self._reclaim_row_btn_selector(device, action)
+        if not present:
+            self.browser.wait_not_present(sel)
+            return
+        self.browser.wait_visible(sel)
+        if disabled:
+            self.browser.wait_js_cond(
```
KKoukiou, line 556: this JS check is more complicated than the xisting with will actually keep working:
        disabled_sel = "[aria-disabled='true']" if disabled else ":not([aria-disabled='true'])"

test/check-storage-reclaim
```diff
@@ -75,7 +75,7 @@ class TestStorageReclaimSpace(VirtInstallMachineCase):
 
         # Test cancelled reclaim is not taking effect
         s.reclaim_remove_device("vda2")
-        s.reclaim_check_action_present("vda2", "delete", True)
+        s.reclaim_check_action_present("vda2", "delete")
```
KKoukiou, line 78: This is an noop change. It's fine to keep it but it's causing unnecessary noise in the commit and makes reviewing take longer. (There are noumerous such changes)

src/components/review/StorageReview.jsx
```diff
@@ -177,6 +180,8 @@ const DeviceRow = ({ disk, isReviewScreen }) => {
                     ...(isReviewScreen ? [{ title: "" }] : []),
                 ],
                 props: {
+                    "data-action": actionType,
```
KKoukiou: If you put `action` var  here instead actionType it will be clear to match this in the tests.

KKoukiou (commented):
I am ok with the PR as is. Because I created it I can't approve it, but you have my ack.

adamkankovsky (approved):

src/components/review/StorageReview.jsx
adamkankovsky: It seems to me that this isn't exactly the kind of situation where this solution would be appropriate

KKoukiou: What I suggest is consistency. If you introduce data-* expressions just handle all attributes like this, otherwise check the text content only.

I suggest this fixup: https://github.com/KKoukiou/anaconda-webui/commit/3d8fdd29d8990b70bb6dcc7327adcdda57a7ade3 

please check, let meknow if you disagree.

adamkankovsky: Yes, this is probably the better option.

test/helpers/review.py
```diff
@@ -55,23 +55,33 @@ def check_disk_row(
...
         disk,
         mount_point="", parent="", size="", reformat="",
-        fs_type=None, is_encrypted=False, rowIndex=None,
+        fs_type=None, is_encrypted=False,
         action="", prefix=""
     ):
-        action = f"format as {fs_type}" if reformat else action or "mount"
+        table = f"{prefix} #storage-review-table-{disk}".strip()
+
+        action_text = f"format as {fs_type}" if reformat else action or "mount"
```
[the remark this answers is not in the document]
KKoukiou: test writers should care to not do that.

PR #1266, opened 2026-04-21
users: implement granular Accounts actions

KKoukiou:
Replace broad setUsersAction usage on the Accounts page with focused action creators. In this way we no longer rebuild the whole users array on every keystroke retriggering effects and causing a render loop.


KKoukiou:
Closing in favor of https://github.com/rhinstaller/anaconda-webui/pull/1268

PR #1268, opened 2026-04-22
users: fix Accounts re-renders

KKoukiou:
`setUsersAction` is now an async thunk that reads the current user list from the backend, merges the first user when needed, copies only defined keys (`assignDefined`), and dispatches when the payload is non-empty.


rvykydal (approved):
Looks good to me.

PR #1274, opened 2026-04-24
[do not merge] PoC for remote installation support

bruno-fs:
## Context

Proof-of-concept for PIN-based authentication as part of the remote installation feature ([INSTALLER-4395](https://redhat.atlassian.net/browse/INSTALLER-4395) / [INSTALLER-3723](https://redhat.atlassian.net/browse/INSTALLER-3723)).

**Do not merge** — keeping this open while the broader design is worked out. It was the basis for the team demo on April 8 and drove the architectural decisions we've aligned on since.

## What This Demonstrates

Cockpit (the framework Anaconda's Web UI is built on) lets you replace its default login page and authentication handler with custom implementations. This PoC exercises both:

### Custom login page

`src/components/login/LoginPage.jsx`, built into `login.html`/`login.js`. A minimal PIN entry form — shows/hides the PIN, submits it as a Basic auth header (PIN as username). Cockpit routes that to the custom auth handler.

Wired up via `[WebService] CustomLoginPage` in `cockpit.conf`.

### Custom auth handler

`src/scripts/cockpit-pin-auth` — a Python script that speaks Cockpit's auth protocol over a Unix socket. Receives the Basic auth credentials from Cockpit, validates the PIN, and if correct, exec's `cockpit.bridge` to start the actual installer session.

Deployed as a socket-activated systemd unit (`cockpit-pin-auth.socket` + `cockpit-pin-auth@.service`), separate from the main cockpit-ws process. Each connection spawns a fresh instance. After the PIN is accepted, Cockpit's session cookie takes over.

### Cockpit config

`src/config/cockpit.conf` sets `CustomLoginPage` and overrides the `[Basic]` auth command. Currently installs to `/etc/cockpit/cockpit.conf`.

**Known issue**: config path should move to an anaconda-owned path (e.g., `/etc/anaconda/cockpit/`) so it doesn't persist to the installed system. Tracked as a follow-up.

## How to Try It

Build an updates image and run it against a Fedora Rawhide boot ISO:

```bash
make create-updates.img
```

Then boot a Rawhide installer with `inst.updates` pointing to the generated `updates.img`, for example with `virt-install`:

```bash
virt-install \
  --location <rawhide-boot-iso> \
  --extra-args "inst.updates=<path-or-url-to-updates.img> inst.webui.remote" \
  ...
```

The PIN is hardcoded to `1234` in `src/scripts/cockpit-pin-auth`. Once the installer is running, point a browser at the VM's IP on port 80 — you'll get the custom login page. Enter `1234` to authenticate.

Session behavior observed in the demo: tab close preserves the session (cookie survives), browser close requires re-login.

## What's Deliberately Incomplete

- **PIN mechanism**: Hardcoded to `1234`. Real implementation would read from a boot parameter and display on the local console.
- **Config path**: Uses `/etc/cockpit/` — should be anaconda-owned.
- **Single-connection enforcement**: Not implemented. Multiple sessions can connect simultaneously (this was live-demoed as a problem — one session can start installation while another changes storage config).
- **TLS**: Not configured. Connection is unencrypted in this PoC.
- **Login page pre-auth info**: IP address / Fedora edition can't easily be shown before login because Cockpit APIs require authentication. Still being figured out.

## What We're Looking For

If you're looking at this from the Cockpit side:
- Is there a simpler way to get OS/network info on the login page without requiring auth?
- What's the recommended mechanism for enforcing a single concurrent session?
- How does Cockpit distinguish a reconnection (same user, dropped connection) from a new session?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,8 +128,17 @@ install: $(DIST_TEST) po/LINGUAS
 	cp browser-ext $(DESTDIR)/usr/libexec/anaconda
 	cp gnome-control-center-ext $(DESTDIR)/usr/libexec/anaconda
 	cp src/scripts/cockpit-coproc-wrapper.sh $(DESTDIR)/usr/libexec/anaconda/
+	cp src/scripts/cockpit-pin-auth $(DESTDIR)/usr/libexec/anaconda/
 	mkdir -p $(DESTDIR)/usr/lib/systemd/system/
 	cp src/systemd/webui-cockpit-ws.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth@.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth.socket $(DESTDIR)/usr/lib/systemd/system/
+	mkdir -p $(DESTDIR)/etc/cockpit
+	cp src/config/cockpit.conf $(DESTDIR)/etc/cockpit/
+	# static files for custom login page must be on this destination 	
+	mkdir -p $(DESTDIR)/usr/share/cockpit/static/
+	ln -sf ../anaconda-webui/login.js $(DESTDIR)/usr/share/cockpit/static/login-anaconda.js
+	ln -sf ../anaconda-webui/login.css $(DESTDIR)/usr/share/cockpit/static/login-anaconda.css
 
 # required for running integration tests;
 TEST_NPMS = \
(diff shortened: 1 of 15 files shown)


bruno-fs:
at this point we can finally close this PR. The core functionality is finally merged on webui and anaconda repos. If you are eager to test, you can try it now with the following steps:

- Clone the official webui repo (https://github.com/rhinstaller/anaconda-webui/)
- Build an updates image and run it against a Fedora Rawhide boot ISO:
```
make create-updates.img
```
- Then boot a Rawhide installer with inst.updates pointing to the generated updates.img, for example with virt-install:
```
virt-install \
  --location <rawhide-boot-iso> \
  --extra-arg inst.updates=<path-or-url-to-updates.img> \
  --extra-arg  inst.webui.remote \
  --extra-arg inst.webui.remote.pin=<your pin here> \
  ...
```

PR #1279, opened 2026-04-28
Fix error dialog flashing on reboot after successful installation

adamkankovsky:
When the user clicks 'Reboot into installed system', exitGui() kills the webui-desktop process, which causes the Anaconda backend to remove the /run/anaconda/backend_ready flag. The file watcher in app.jsx detects this removal while the UI is still alive and triggers the "The Anaconda installation has stopped unexpectedly" error dialog.

Call Boss.Quit() via D-Bus before killing the process for a clean shutdown, and track the intentional exit so the backend_ready watcher skips the error when the user initiated the quit.

### DIFF
--- a/src/helpers/exit.js
+++ b/src/helpers/exit.js
@@ -6,9 +6,16 @@ import cockpit from "cockpit";
 
 import { debug, error } from "./log.js";
 
+let _isExiting = false;
+
+export const isExiting = () => _isExiting;
+
 export const exitGui = () => {
+    _isExiting = true;
+
     const pidFile = cockpit.file("/run/anaconda/webui_script.pid", { superuser: "try" });
     let pid;
+
     pidFile.read()
             .then(content => {
                 pid = content.trim();
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -12,6 +12,7 @@ import { clients } from "../apis/index.js";
 import { initialState, reducer, useReducerWithThunk } from "../reducer.js";
 
 import { getInstallerConfValue, parseAnacondaConfBool, readConf } from "../helpers/conf.js";
+import { isExiting } from "../helpers/exit.js";
 import { debug } from "../helpers/log.js";
 import { getAnacondaUIVersion, getAnacondaVersion } from "../helpers/product.js";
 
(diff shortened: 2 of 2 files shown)


src/helpers/exit.js
```diff
@@ -4,12 +4,23 @@
...
 
+import { quit } from "../apis/boss.js";
+
 import { debug, error } from "./log.js";
 
+let _isExiting = false;
+
+export const isExiting = () => _isExiting;
+
 export const exitGui = () => {
```
KKoukiou, line 13: I am quite puzzled on why we do this. Stopping the service or similarly stopping the modules should be an internal handling done by backend. Here we should just call some exit method provided by a backend API. 

I see there is one method in runtime module called Exit that you introduced recently. Cant' we use that?

adamkankovsky: I apologize; two changes got mixed up here a bit—I probably should have split them into two commits. 
1. I have properly shut down DBus and all its modules before terminating the installation process itself. 
This doesn’t directly affect it, but it also impacts this https://github.com/rhinstaller/anaconda-webui/pull/1279/changes#diff-09729c32a5f779fcf379a03c27ad88998390f203530749ffa1c88fdfed364f1eL160 listener , which checks whether DBus is connected. That’s why we need to add this option for a complete shutdown—so we don’t mind if DBus disconnects and we don’t need to display an error.

adamkankovsky: In theory, I could try using `exit`, but I'm afraid it might not work properly with the web UI, and the browser or something like that might keep running. But it's true that, in theory, it could remove the unnecessary code related to storing the PID.

adamkankovsky: So I overlooked the fact that "exit" is only for KS commands like "reboot" and "shutdown."

adamkankovsky:
This is the only correct approach. While the `quit` command will terminate the Boss module and all other modules, the daemon will restart them. It is not worth calling `exit` because webui-desktop now controls the installation process, so it must be terminated first.

KKoukiou (approved):
I have trouble testing that because i can't boot the ISO today for some issue unfortunately, but the PR is clean. I trust you tested it.

PR #1281, opened 2026-04-29
Extract per-page review summary to the relevant component directory

KKoukiou:
Ok now it looks like this:
Screenshot 2026-04-30 at 08-18-26 Anaconda Web UI


rvykydal (approved):
Looks good to me.
Minor things:
2nd commit seems to have wrong commit message (localization -> datetime).
3rd commit also adds incompleteness indicator to installation method if I am reading it well.

KKoukiou:
> Looks good to me. Minor things: 2nd commit seems to have wrong commit message (localization -> datetime). 3rd commit also adds incompleteness indicator to installation method if I am reading it well.

good catch, 3rd commit is not ok.

KKoukiou:
Fixed the comments from Radek above, I will merge to not block it further as people are on PTO; if something else appears let's do followup.

PR #1288, opened 2026-05-05
tests: specify VM memory also when spawning the VM with webui_testvm

KKoukiou:
This is used for running a test vm locally, not CI.


rvykydal (approved):
Looks good to me.

bruno-fs (approved):

PR #1290, opened 2026-05-06
Extract extlink URL handling into shared helper

KKoukiou:
Move convertToExtlinkIfNeeded to a dedicated extlink helper module and reuse it in all flows so external link behavior is centralized.


adamkankovsky (approved):
LGTM, thanks

PR #1291, opened 2026-05-06
Fix auto-reboot/poweroff after kickstart installation

adamkankovsky:
When a kickstart file specifies a reboot or poweroff command, the WebUI now automatically triggers the corresponding action after installation completes, matching the behavior of the GUI and TUI frontends.

After the installation task succeeds in automated mode, the progress screen reads the RebootData.action property from the Runtime D-Bus module. If KS_REBOOT (1) or KS_SHUTDOWN (2) is set, exitGui() is called automatically without waiting for user interaction.

Previously, the WebUI always showed the "Reboot" button and waited for a manual click, even when a kickstart reboot/poweroff command was present.

### DIFF
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -23,6 +23,7 @@ import { OsReleaseContext, SystemTypeContext } from "../../contexts/Common.jsx";
 import { EmptyStatePanel } from "cockpit-components-empty-state.jsx";
 
 import { Feedback } from "./Feedback.jsx";
+import { useAutoReboot } from "./useAutoReboot.js";
 
 import "./InstallationProgress.scss";
 
@@ -38,7 +39,7 @@ const progressStepsMap = {
     SYSTEM_CONFIGURATION: 3,
 };
 
-export const InstallationProgress = ({ onCritFail }) => {
+export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
     const [status, setStatus] = useState();
     const [statusMessage, setStatusMessage] = useState("");
     const [steps, setSteps] = useState();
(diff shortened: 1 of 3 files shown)


src/components/installation/InstallationProgress.jsx
```diff
@@ -38,12 +43,13 @@ const progressStepsMap = {
...
 };
 
-export const InstallationProgress = ({ onCritFail }) => {
+export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
     const [status, setStatus] = useState();
     const [statusMessage, setStatusMessage] = useState("");
     const [steps, setSteps] = useState();
     const [currentProgressStep, setCurrentProgressStep] = useState(0);
     const refStatusMessage = useRef("");
+    const refAutomatedInstall = useRef(automatedInstall);
```
KKoukiou: Why ref? automatedInstall is just a bool,. it wont change mid install which I guess what this code prevents.

src/components/installation/InstallationProgress.jsx
```diff
@@ -93,6 +99,18 @@ export const InstallationProgress = ({ onCritFail }) => {
                         taskProxy.addEventListener("Succeeded", () => {
                             setStatus("success");
                             setCurrentProgressStep(4);
+
```
KKoukiou: I prefer this code exists as a seperate hook, useAutoProgress that will take as input status var, and once it's success it can do this.

src/components/installation/InstallationProgress.jsx
```diff
@@ -93,6 +99,18 @@ export const InstallationProgress = ({ onCritFail }) => {
...
+                            if (refAutomatedInstall.current) {
+                                getRebootData()
+                                        .then(rebootData => {
```
KKoukiou: please use async await.

adamkankovsky:
Tested and works

src/components/installation/InstallationProgress.jsx
```diff
@@ -38,7 +43,7 @@ const progressStepsMap = {
     SYSTEM_CONFIGURATION: 3,
 };
 
-export const InstallationProgress = ({ onCritFail }) => {
+export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
```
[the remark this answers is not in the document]
KKoukiou: It's passed through componentProps.

KKoukiou (approved):
the TUI and GUI handle this the exact same way

PR #1292, opened 2026-05-06
storage: avoid UnknownDeviceError after disk removal and rescan

KKoukiou:
After a disk was removed from the VM and the user rescanned, D-Bus SelectedDisks could still list that device name for a while. The storage module's GetDiskTotalSpace, GetDiskFreeSpace, and GetMountPointConstraints resolve names against the current device tree only, so a stale id raised UnknownDeviceError (rhbz#2456923).

We now filter selectedDisks by usableDisks before calling any APIs. Usable disks are derived from the same module state as the tree after a scan, so a removed disk cannot be in usableDisks and is dropped before any viewer call.

Resolves: rhbz#2456923

### DIFF
--- a/src/helpers/storage.js
+++ b/src/helpers/storage.js
@@ -323,3 +323,18 @@ export const selectDefaultDisks = ({ ignoredDisks, selectedDisks, usableDisks })
     }
     return [];
 };
+
+/**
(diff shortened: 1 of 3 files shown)


adamkankovsky (approved):
The test does a good job of verifying functionality, and it seems fine to me too

PR #1298, opened 2026-05-12
test: fix pretend_live_workstation_iso anaconda.conf sed

KKoukiou:
The pattern /[anaconda]/ is a sed character class, not the [Anaconda] header, so hidden_webui_pages was appended after many lines and duplicated keys broke configparser when anaconda reloaded the file.

Write the option once under [User Interface].


rvykydal (approved):
Looks good to me

rvykydal (approved):

PR #1299, opened 2026-05-12
test: Restore VirtMachine boot console capture for ISO-based installer VMs

KKoukiou:
The anaconda-webui test suite subclasses VirtMachine to start guests from an installer ISO via virt-install, whereas Cockpit's VirtMachine boots from a prepared qcow image. That split meant we don't support all features VirtMachine does.

Here we re-implement serial console capture into console_file (boot logs for debugging). In order to do that, adjust virt-install to write console logs to a file, and also set the _domain attribute so that we can leave the wait_poweroff() and kill() method to be done by the parent class's code that deals with the logging as well.

### DIFF
--- a/test/machine_install.py
+++ b/test/machine_install.py
@@ -19,6 +19,7 @@ BOTS_DIR = f'{ROOT_DIR}/bots'
 sys.path.append(BOTS_DIR)
 
 # pylint: disable=import-error
+import libvirt  # type: ignore[import-untyped]
 from machine.machine_core import timeout
 from machine.testvm import (
     Machine,  # nopep8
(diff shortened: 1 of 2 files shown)


test/machine_install.py
```diff
@@ -167,9 +196,10 @@ def start(self):
                 f"--os-variant=detect=on "
                 f"--memory {self.memory_mb} "
                 "--noautoconsole "
+                f"{serial_opt}"
                 f"--graphics vnc,listen={self.ssh_address} "
                 "--extra-args "
-                f"'inst.sshd inst.webui.remote {selinux}{inst_ks_arg} "
+                f"'{virt_kargs}{inst_graphical}inst.sshd inst.webui inst.webui.remote {selinux}{inst_ks_arg} "
```
rvykydal, line 202: I guess we are adding inst.webui to override inst.grapical?

rvykydal (approved):
Looks good to me.

test/machine_install.py
KKoukiou: inst.graphical  does not imply GTK. When I pass the console in the boot options, which is needed for logging the boot logs. the installer auto-switches to text mode, unless I pass the inst.graphical. So this is in addition to inst.webui

(Maybe inst.webui is not even needed I now realize as it's implied when it's installed)

PR #1311, opened 2026-05-20
storage: show specific error when /boot/efi is on vfat without ESP

adamkankovsky:
vfat partitions without the EFI System type are reported as vfat, not efi. Show a clearer validation message instead of requiring format type efi.


KKoukiou (approved):
I split some logic into helpers/storage (because I am not sure if you are working these days) and approved. thanks

PR #1313, opened 2026-05-21
Show reused efi correctly in review

rvykydal:


src/components/storage/StorageReview.jsx
```diff
@@ -90,7 +91,8 @@ const DeviceRow = ({ disk, isReviewScreen }) => {
         const size = cockpit.format_bytes(devices[device].size.v);
         const request = requests.find(request => request["device-spec"] === device);
         let format = devices[device].formatData.type.v;
-        const isReformattedMountPoint = (!request && !reusedMountPoints?.includes(mount)) || request?.reformat;
+        const isImplicitlyReusedMountPoint = originalDevices[device] && !reformattedMountPoints?.includes(mount);
```
KKoukiou: This is smart to check it like this. But can't we rather have the reused-mount-points in backend containing /boot/efi?

rvykydal: Interesting idea. I am quite afraid to start using the attribute used now for home reuse also for regular autopartitioning but I will take a look. My concern is if we use it to track /boot/efi as reused we should also set it as such (where? frontend / backend by default of from config?) and use it to reuse /boot/efi in normal autopart. But maybe it is rather a way to unify things then to break them.

rvykydal:
Also the tests are still failing, I'll switch to ready for review when they pass.

rvykydal:
> Also the tests are still failing, I'll switch to ready for review when they pass.

Fixed the failing tests with PR update.
I will still check the reused_mount_points approach.

rvykydal:
> > Also the tests are still failing, I'll switch to ready for review when they pass.
> 
> Fixed the failing tests with PR update. I will still check the reused_mount_points approach.

@KKoukiou So looking more into it confirms my doubts from the [comment above](https://github.com/rhinstaller/anaconda-webui/pull/1313#discussion_r3296877976):

> Interesting idea. I am quite afraid to start using the attribute used now for home reuse also for regular autopartitioning but I will take a look. My concern is if we use it to track /boot/efi as reused we should also set it as such (where? frontend / backend by default of from config?) and use it to reuse /boot/efi in normal autopart. But maybe it is rather a way to unify things then to break them.

and it does not make sense to me to use `reused_mount_point` property of [PartitioningRequest](https://github.com/rhinstaller/anaconda/blob/a47b1162238cdccb15f34f7aff6d44348dc57245/pyanaconda/modules/common/structures/partitioning.py#L28) for this purpose. (contain `"/boot/efi"` for use-free-space autopartitioning)
1) It is a request (not state holder). Something needs to set it up (request `/boot/efi` to be reused) - would it be frontend when creating autopart request? Or would it be some backend default or configuration setting?
2) The logic for reusing `/boot/efi` is in [pyanaconda/modules/storage/partitioning/automatic/utils.py.](https://github.com/rhinstaller/anaconda/blob/a47b1162238cdccb15f34f7aff6d44348dc57245/pyanaconda/modules/storage/partitioning/automatic/utils.py#L373) It is using bootloader module and its validation functionality to search for the bootlodader stage1 device that would potentialy be reused. The request does not have to be fulfiled (existing deivce valid as stage1 target is not found). Then the request can't be used in the frontend [review screen](https://github.com/rhinstaller/anaconda-webui/blob/9ebd95780b86d16dbc4df87f179a959bc6eb7937/src/components/storage/StorageReview.jsx#L100) to determie if `"/boot/efi"` is reused or not.
3) The format/mount action of `"/boot/efi"` needs to be determined dynamically - as in current version of the PR.

KKoukiou (approved):
Please squash the commits, the second is fixing a thing introduced in the first. You can enable squash and merge option in the settings in order to not repush

PR #1317, opened 2026-05-26
tests: adjust scenario trigger for compose testing

KKoukiou:
There is not `efi` scenario anymore, UEFI firmware test runs are now done by default and the extra scenario is `bios`.


rvykydal (approved):

M4rtinK (approved):

PR #1319, opened 2026-05-27
storage: show sizes in binary units and align tests

KKoukiou:
Add formatBytes() with base2 sizing matching blivet and Anaconda Backend. Update integration test expectations to MiB/GiB and use  EFI_PARTITION_DEFAULT_SIZE_SPACE (500 MiB) for default EFI rows.
Note: cockpit-storaged still uses decimal units

### DIFF
--- a/src/helpers/storage.js
+++ b/src/helpers/storage.js
@@ -182,14 +182,33 @@ export const getDeviceTypeInfo = (device) => {
     return deviceType;
 };
 
+const isPlainObject = (value) => value !== null && typeof value === "object" && !Array.isArray(value);
+
+/**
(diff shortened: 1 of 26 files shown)


src/helpers/storage.js
```diff
@@ -182,14 +182,33 @@ export const getDeviceTypeInfo = (device) => {
     return deviceType;
 };
 
+const isPlainObject = (value) => value !== null && typeof value === "object" && !Array.isArray(value);
+
+/**
+ * Format a byte count (binary KiB/MiB/GiB).
```
tomasfratrik, line 188: What about TiB ? it is in the multiplier

PR #1347, opened 2026-06-24
Fix automated multi-disk kickstart installs by selecting disks on review

KKoukiou:


KKoukiou:
Let's do that in backend.

PR #1349, opened 2026-06-25
webui-desktop: fallback to Firefox when configured browser is unavailable

xasciii:
While installing Fedora from the live environment I encountered an issue where the installer failed to launch because the configured browser command was unavailable This change preserves the configured browser when it is valid and falls back to Firefox when necessary, avoiding the need for a manual workaround


KKoukiou (changes requested):
The default browser is configured via the profiles files per distro: 
https://github.com/rhinstaller/anaconda/blob/main/data/profile.d/fedora.conf#L25

Default is slitherer, Workstation uses firefox.
For the spin you are tagetting just make sure to set the default browser there. 

firefox does  not exist in all spins so this workaround is not really global. Thanks.

KKoukiou:
Closing - see review for justification.

PR #1357, opened 2026-07-07
webui: handle KeyboardConfigurationError via step notification

KKoukiou:
When the backend raises KeyboardConfigurationError (e.g., live system has only non-XKB layouts like IBus 'anthy' or 'libpinyin'), display it as a step notification on the language/localization screen.

User experience:
- Japanese user with only 'anthy' configured sees inline alert: "The live system has layout 'anthy' which can't be used for installation."
- User can click "Change system keyboard layout" to configure manually

Related: rhbz#2496409
Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

Screenshot 2026-07-07 at 13-53-18 Virtual machines - kkoukiou@fedora

### DIFF
--- a/src/actions/localization-actions.js
+++ b/src/actions/localization-actions.js
@@ -72,12 +72,18 @@ export const getKeyboardLayoutsAction = () => {
     };
 };
 
-export const getKeyboardConfigurationAction = () => {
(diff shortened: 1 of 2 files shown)


M4rtinK (approved):

PR #1359, opened 2026-07-08
Reconnect to active installation task after page reload

adamkankovsky:
Use the Boss ActiveInstallationTask API to resume ongoing installations when the UI is reloaded, and add a test for the reconnect flow.

### DIFF
--- a/src/apis/boss.js
+++ b/src/apis/boss.js
@@ -6,7 +6,7 @@
 import cockpit from "cockpit";
 
 import { error } from "../helpers/log.js";
-import { _callClient } from "./helpers.js";
+import { _callClient, _getProperty } from "./helpers.js";
 
 const OBJECT_PATH = "/org/fedoraproject/Anaconda/Boss";
 const INTERFACE_NAME = "org.fedoraproject.Anaconda.Boss";
@@ -55,6 +55,13 @@ export const getSteps = ({ task }) => {
         ["org.fedoraproject.Anaconda.Task", "Steps"]
     )
             .then(ret => ret[0]);
+};
+
+/**
+ * @returns {Promise}           Resolves the object path of the active installation task, or ""
+ */
+export const getActiveInstallationTask = () => {
+    return _getProperty(BossClient, OBJECT_PATH, INTERFACE_NAME, "ActiveInstallationTask");
 };
 
 /**
--- a/src/components/AnacondaWizard.jsx
+++ b/src/components/AnacondaWizard.jsx
@@ -10,6 +10,8 @@ import React, { useContext, useEffect, useRef, useState } from "react";
 import { PageSection, PageSectionTypes } from "@patternfly/react-core/dist/esm/components/Page/index.js";
 import { Wizard, WizardStep } from "@patternfly/react-core/dist/esm/components/Wizard/index.js";
 
+import { getActiveInstallationTask } from "../apis/boss.js";
+
 import { PageContext, PayloadContext, StorageContext, SystemTypeContext, UserInterfaceContext } from "../contexts/Common.jsx";
 
 import { AnacondaPage } from "./AnacondaPage.jsx";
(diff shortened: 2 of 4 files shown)


src/apis/boss.js
```diff
@@ -57,6 +57,17 @@ export const getSteps = ({ task }) => {
             .then(ret => ret[0]);
 };
 
+/**
+ * @returns {Promise}           Resolves the object path of the active installation task, or ""
+ */
+export const getActiveInstallationTask = () => {
+    return _getProperty(BossClient, OBJECT_PATH, INTERFACE_NAME, "ActiveInstallationTask")
+            .catch(err => {
```
KKoukiou: I don't se why are ignoring exceptions here. If there is no ActiveInstallation task backend returns "". If this fails it means that something went wrong and we should show bug report dialog.

src/components/AnacondaWizard.jsx
```diff
@@ -65,6 +67,21 @@ export const AnacondaWizard = ({ automatedInstall, currentStepId, dispatch, isFe
         }
     }, [currentStepId, firstStepId, path, setCurrentStepId]);
 
+    const finalStepId = stepsOrder[stepsOrder.length - 1]?.id;
```
KKoukiou, line 70: I don't think we should be checking step ID here. If there is installation running always show the last page.

test/check-progress
```diff
@@ -65,6 +65,49 @@ class TestInstallationProgress(VirtInstallMachineCase):
...
+        i.open()
+
+        i.reach(i.steps.REVIEW)
+        i.begin_installation(needs_confirmation=False)
+
+        with b.wait_timeout(300):
+            b.wait_in_text("h2", "Installing")
+
+        # Reload the page while installation is running to simulate reconnect
+        b.open("/cockpit/@localhost/anaconda-webui/index.html")
```
KKoukiou: does this reload? 

you can really reload with: b.reload();

test/check-progress
```diff
@@ -65,6 +65,49 @@ class TestInstallationProgress(VirtInstallMachineCase):
...
+
+        # The UI should auto-navigate to progress and reconnect to
+        # the active installation task
+        with b.wait_timeout(300):
```
KKoukiou: I don't thin we need 5 misn timeout for this. It should bve instant.

test/check-progress
```diff
@@ -65,6 +65,49 @@ class TestInstallationProgress(VirtInstallMachineCase):
...
+            b.wait_visible(".pf-v6-c-progress-stepper")
+            b.wait_in_text("h2", "Installing")
+
+        p.wait_done()
```
KKoukiou: For making the test faster we do't need to finish the installation. I would remove everything after this line. we just want to test it can reconect.

KKoukiou (changes requested):

test/check-progress
```diff
@@ -65,6 +65,39 @@ class TestInstallationProgress(VirtInstallMachineCase):
...
+        """
+        Description:
+            Verify that the UI reconnects to an ongoing installation after a page reload.
+
+        Expected results:
+            - After beginning installation and reloading the page, the UI
+              auto-navigates to the progress screen
+        """
+
+        # HACK Ignore some selinux errors
```
KKoukiou: We dont deal with selinux in webui tests. That's why we run with TEST_AUDIT_NO_SELINUX in CI and locally always

test/check-progress
```diff
@@ -65,6 +65,39 @@ class TestInstallationProgress(VirtInstallMachineCase):
 
         self.handleReboot()
 
+    def testReconnect(self):
```
KKoukiou: Instead of adding a new test c an you just add a reload check in testBasic in check-progress?

src/components/AnacondaWizard.jsx
```diff
@@ -65,6 +67,21 @@ export const AnacondaWizard = ({ automatedInstall, currentStepId, dispatch, isFe
...
+            return;
+        }
+        getActiveInstallationTask()
+                .then(activeTask => {
+                    checkedActiveTaskRef.current = true;
+                    if (activeTask) {
+                        cockpit.location.go([finalStepId]);
+                    }
+                });
+    }, [finalStepId, path]);
```
KKoukiou: Path is not used. Why you need it in the dependency array? In fact I think that if you remove 'path' this can entrer only once, because finalStepId does not change, so you can also remove the logic around checkedActiveTaskRef'. (drop checkedActiveTaskRef completely)

test/check-progress
```diff
@@ -41,6 +39,13 @@ class TestInstallationProgress(VirtInstallMachineCase):
...
         with b.wait_timeout(300):
             b.wait_in_text("h2", "Installing")
+
+        # Reload the page while installation is running to verify reconnect
+        b.reload()
+        b.wait_visible(".pf-v6-c-progress-stepper")
+        b.wait_in_text("h2", "Installing")
+
+        with b.wait_timeout(300):
             b.wait_in_text(".pf-v6-c-empty-state", "Storage configuration: Storage is currently being configured.")
```
KKoukiou: You can remove these two. We don't care about specifc prgress messages - they are fragile to check.

KKoukiou (approved):

PR #1360, opened 2026-07-08
Allow automated installations to proceed despite storage warnings

KKoukiou:
Storage warnings (like boot partition smaller than recommended) should not block automated installations from auto-proceeding through the review screen. This matches GTK/TUI behavior where only incomplete pages block auto-proceed.


M4rtinK (approved):
Looks good to me. :)

KKoukiou:
The test failure is unrelated. Because the test file was affected it forced everything to run 3 times uncovering flakiness. Let's address this separately.

PR #1367, opened 2026-07-13
Add Restart button to post-installation screen for live media

KKoukiou:
On live media, users only saw "Exit to live desktop" after installation, but were told to reboot. Show also a "Reboot to installed system" button that sets the Reboot D-Bus property on the Runtime module and exits through anaconda's normal exit path, ensuring proper cleanup.

Resolves: rhbz#2326672

Screenshot 2026-07-13 at 17-41-41 Virtual machines - kkoukiou@fedora

Needs: https://github.com/rhinstaller/anaconda/pull/7176

### DIFF
--- a/src/apis/runtime.js
+++ b/src/apis/runtime.js
@@ -8,7 +8,7 @@ import cockpit from "cockpit";
 import { getPasswordPoliciesAction } from "../actions/runtime-actions.js";
 
 import { debug, error } from "../helpers/log.js";
-import { _getProperty } from "./helpers.js";
+import { _getProperty, _setProperty } from "./helpers.js";
 
 const OBJECT_PATH = "/org/fedoraproject/Anaconda/Modules/Runtime/UserInterface";
 const INTERFACE_NAME = "org.fedoraproject.Anaconda.Modules.Runtime.UserInterface";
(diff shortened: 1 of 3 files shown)


M4rtinK (approved):

AdamWill:
"Reboot to installed system" seems like a promise we can't really make.

In ye olde, olde, olde, OLDE days, the installer ejected the CD (you install from a CD, right? A CD? It's kinda like a drink coaster only shinier) when you rebooted, to try and ensure we'd boot to the installed system on next boot. But these days nobody installs from optical media, I don't know if webui even tries to eject it, and we can't eject a USB stick and we don't know exactly how the user's firmware is configured. So we don't really know what the hell the system will boot to.

It might be safer to just label the button "Reboot".

KKoukiou:
> "Reboot to installed system" seems like a promise we can't really make.
> 
> In ye olde, olde, olde, OLDE days, the installer ejected the CD (you install from a CD, right? A CD? It's kinda like a drink coaster only shinier) when you rebooted, to try and ensure we'd boot to the installed system on next boot. But these days nobody installs from optical media, I don't know if webui even tries to eject it, and we can't eject a USB stick and we don't know exactly how the user's firmware is configured. So we don't really know what the hell the system will boot to.
> 
> It might be safer to just label the button "Reboot".

agreed, I will adjust. TBH I am always testing on VMs and virt-install does the ejection of the installation media for us so I did not even hit this.

KKoukiou:
'It's kinda like a drink coaster only shinier' :rofl:

PR #1373, opened 2026-07-16
storage: Add support for home reuse partitioning with Stratis

vojtechtrefny:
Unfortunately home reuse currently works only with default partitioning so to make it work with stratis, it needs to be changed in the config.


rvykydal (approved):
Looks good to me, thank you.

PR #1376, opened 2026-07-21
Encrypted Stratis support -- mount point assignment, frontend

vojtechtrefny:
Needs https://github.com/rhinstaller/anaconda/pull/7191 using encrypted pools created in cockpit needs https://github.com/cockpit-project/cockpit/pull/23538

The biggest part of this is changing check for "is device encrypted" from checking if format is LUKS to checking if format is LUKS or a device is encrypted stratis pool. Most of the code changes is then renaming functions and properties from `*luks*` to `*encrypted*`.


rvykydal (approved):
Looks good to me.

KKoukiou (approved):
Awesome work.

PR #1377, opened 2026-07-21
Check InstallationFinished before starting new install

bruno-fs:
When a user refreshes the browser after a successful installation, the WebUI could not distinguish between 'never started' and 'finished successfully' because ActiveInstallationTask is cleared on completion. This caused a new install attempt that immediately failed.

Check the new InstallationFinished D-Bus property (INSTALLER-4824) as a fallback when no active task exists. If the installation already finished, navigate to the success screen instead of starting a new installation.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/src/apis/boss.js
+++ b/src/apis/boss.js
@@ -65,6 +65,13 @@ export const getActiveInstallationTask = () => {
 };
 
 /**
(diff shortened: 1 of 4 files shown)


bruno-fs:
depends on rhinstaller/anaconda#7192

bruno-fs:
dropped in favor of #1404

PR #1382, opened 2026-07-23
webui_testvm: Add --kickstart and --pause-at-summary CLI options

bruno-fs:
Pass kickstart file name and pause-at-summary flag through to VirtInstallMachine, enabling automated kickstart installs from the test VM launcher.

### DIFF
--- a/test/webui_testvm.py
+++ b/test/webui_testvm.py
@@ -18,11 +18,16 @@ def cmd_cli():
     parser.add_argument("--rsync", help="Rsync development files over on startup", action='store_true')
     parser.add_argument("--host", help="Hostname to rsync", default='test-updates')
     parser.add_argument("--bios", help="Start the VM with BIOS firmware (UEFI is the default)", action='store_true')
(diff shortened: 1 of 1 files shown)


KKoukiou (approved):
How about we introduce one single option called --boot

M4rtinK (approved):
Looks good to me, thanks! :)

M4rtinK:
> How about we introduce one single option called --boot

I'm not sure I follow. :)

Also I think both options introduced in this PR makes sense separately as well as when combined. You might want to just use a kickstart, other time you want to be sure automated installation does not start with kickstart.

I guess just `--pause-at-summary` with no kickstart will be kind of a no-op, but I think that is fine. :)

bruno-fs:
> How about we introduce one single option called --boot

What do you have in mind for this option?

KKoukiou:
> > How about we introduce one single option called --boot
> 
> What do you have in mind for this option?

Sorry i started writing a review - then thought I removed it - but it's here. Basically I was trying to propose a single boot options parameter- where one could override any boot option. But so far - with few boot options being passed l like this solution better.

PR #1384, opened 2026-07-24
Dnf install source minimal (HTTP/HTTPS + baseurl, mirrorlist and metalink)

tomasfratrik:
Adds a minimal DNF installation source spoke: closest mirror and http/https repositories (base URL, mirrorlist, and metalink). Values set via kickstart (`url --url`, `--metalink`, `--mirrorlist`) are shown pre-filled in the spoke and can still be changed.

Commits:
- Add dnf payload to test/webui_testvm.py
- Add payload source Redux state and D-Bus API
- Add installation source spoke for DNF payloads
- Refresh software selection after installation source changes
- test: check installation source is not shown on Live ISO review step

If you enter an invalid URL, the UI waits on backend setup with no way to cancel — a cancel button might be worth adding later.

Resolves: INSTALLER-3621
Assisted-by: Composer 2.5

Screenshot From 2026-07-24 11-27-05
Screenshot From 2026-07-24 11-27-17

### DIFF
--- a/src/actions/payload-dnf-actions.js
+++ b/src/actions/payload-dnf-actions.js
@@ -4,32 +4,40 @@
  */
 
 import {
+    getDefaultEnvironment,
     getEnvironmentData,
     getEnvironments,
     getGroupData,
     getPackagesKickstarted,
     getPackagesSelection,
+    resolveEnvironment,
+    setPackagesSelection,
 } from "../apis/payload_dnf.js";
 
 export const getPayloadEnvironmentsAction = () => {
     return async (dispatch) => {
         const environmentIds = await getEnvironments();
 
-        // Fetch environment data for each environment to get descriptions
-        const environmentDataPromises = environmentIds.map(async (envId) => {
-            const envData = await getEnvironmentData(envId);
-            return {
-                description: envData.description,
-                id: envId,
-                name: envData.name,
-            };
-        });
+        const environmentResults = await Promise.all(
+            environmentIds.map(async (envId) => {
+                try {
+                    const envData = await getEnvironmentData(envId);
+                    return {
+                        description: envData.description,
+                        id: envId,
+                        name: envData.name,
+                    };
+                } catch {
+                    return null;
+                }
+            })
+        );
 
-        const environments = await Promise.all(environmentDataPromises);
+        const environments = environmentResults.filter(Boolean);
 
         return dispatch({
             payload: { environments },
-            type: "SET_PAYLOAD_ENVIRONMENTS"
+            type: "SET_PAYLOAD_ENVIRONMENTS",
         });
     };
 };
@@ -76,7 +84,53 @@ export const getPayloadGroupsAction = (environment) => {
 
         return dispatch({
             payload: { groups },
-            type: "SET_PAYLOAD_GROUPS"
+            type: "SET_PAYLOAD_GROUPS",
         });
+    };
+};
+
+/**
+ * Reload environments and re-validate the current selection after the source changes.
+ */
+export const refreshPayloadSoftwareSelectionAction = () => {
+    return async (dispatch) => {
+        await dispatch(getPayloadEnvironmentsAction());
+
+        let selection = await getPackagesSelection();
+        let environment = selection?.environment;
+
+        if (environment) {
+            const resolved = await resolveEnvironment(environment);
+            if (!resolved) {
+                environment = await getDefaultEnvironment();
+                if (environment && await resolveEnvironment(environment)) {
+                    await setPackagesSelection({ environment, groups: [] });
+                } else {
+                    await setPackagesSelection({ environment: "", groups: [] });
+                    environment = "";
+                }
+                await dispatch(getPayloadPackagesSelectionAction());
+                selection = await getPackagesSelection();
+                environment = selection?.environment;
+            }
+        }
+
+        if (environment) {
+            try {
+                await dispatch(getPayloadGroupsAction(environment));
+            } catch {
+                dispatch({
+                    payload: { groups: [] },
+                    type: "SET_PAYLOAD_GROUPS",
+                });
+            }
+        } else {
+            dispatch({
+                payload: { groups: [] },
+                type: "SET_PAYLOAD_GROUPS",
+            });
+        }
+
+        await dispatch(getPayloadPackagesSelectionAction());
     };
 };
--- a/src/apis/payload_source.js
+++ b/src/apis/payload_source.js
@@ -1,0 +1,186 @@
+/*
+ * Copyright (C) 2026 Red Hat, Inc.
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+
+import cockpit from "cockpit";
+
+import { _getProperty, objectFromDbus, objectToDbus } from "./helpers.js";
+
+import { PayloadDNFClient } from "./payload_dnf.js";
+import { PayloadsClient } from "./payloads.js";
+
+const PAYLOADS_OBJECT_PATH = "/org/fedoraproject/Anaconda/Modules/Payloads";
+const PAYLOADS_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads";
+const PAYLOAD_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Payload";
+const SOURCE_BASE_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Source";
+const SOURCE_REPOSITORY_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Source.Repository";
+const SOURCE_CLOSEST_MIRROR_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Source.ClosestMirror";
+const SOURCE_CDROM_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Source.CDROM";
+const SOURCE_REPO_PATH_INTERFACE = "org.fedoraproject.Anaconda.Modules.Payloads.Source.RepoPath";
+
+const getClient = () => new PayloadsClient().client;
+
+const getPayloadPath = () => PayloadDNFClient.instance.payload;
+
+/**
+ * Create a new source of the given type.
+ * @param {string} sourceType - e.g. "URL"
+ * @returns {Promise<string>} D-Bus object path of the created source
+ */
+export const createSource = async (sourceType) => {
+    return getClient().call(
+        PAYLOADS_OBJECT_PATH, PAYLOADS_INTERFACE, "CreateSource", [sourceType]
+    )
+            .then(res => res[0]);
+};
+
+/**
+ * Get the list of source object paths attached to the current payload.
+ * @returns {Promise<string[]>}
+ */
+export const getPayloadSources = async () => {
+    return _getProperty(
+        PayloadsClient, getPayloadPath(), PAYLOAD_INTERFACE, "Sources"
+    );
+};
+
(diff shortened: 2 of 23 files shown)


src/apis/payload_source.js
```diff
@@ -0,0 +1,186 @@
...
+export const runPayloadTask = ({ task }) => {
+    return new Promise((resolve, reject) => {
+        let succeededEmitted = false;
+        const taskProxy = getClient().proxy(
+            "org.fedoraproject.Anaconda.Task",
+            task
+        );
+        const addEventListeners = () => {
+            taskProxy.addEventListener("Stopped", () => taskProxy.Finish().catch(reject));
+            taskProxy.addEventListener("Succeeded", () => {
```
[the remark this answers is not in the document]
tomasfratrik: runPayloadTask intentionally mirrors runStorageTask in src/apis/storage.js.

src/components/software/SoftwareSelection.jsx
```diff
@@ -167,38 +166,195 @@ const GroupPackagesSelection = () => {
...
+const SoftwareSelectionFooter = ({ applyCurrentSelection }) => {
+    const { setIsFormDisabled } = useContext(PageContext) ?? {};
+
+    const onNext = async ({ goToNextStep }) => {
+        setIsFormDisabled(true);
+        try {
+            await applyCurrentSelection();
+            goToNextStep();
+        } catch {
+            // Error alert is shown by applyCurrentSelection.
```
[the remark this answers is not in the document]
tomasfratrik: Makes sense I think, I will fix this

src/components/software/SoftwareSelection.jsx
```diff
@@ -5,34 +5,41 @@
...
+    resolveEnvironment,
+    setPackagesSelection,
+} from "../../apis/payload_dnf.js";
+
+import {
+    getPayloadGroupsAction,
+    refreshPayloadSoftwareSelectionAction,
+} from "../../actions/payload-dnf-actions.js";
 
 import { PageContext, PayloadContext } from "../../contexts/Common.jsx";
```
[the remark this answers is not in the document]
tomasfratrik: I guess this is not bad idea either, so I will do it

src/components/source/InstallationSource.jsx
```diff
@@ -0,0 +1,321 @@
...
+    const { source } = useContext(PayloadContext);
+
+    const [protocol, setProtocol] = useState(PROTOCOL_CLOSEST_MIRROR);
+    const [url, setUrl] = useState("");
+    const [urlType, setUrlType] = useState("BASEURL");
+
+    const [error, setError] = useState(null);
+    const [initialized, setInitialized] = useState(false);
+    const [isApplying, setIsApplying] = useState(false);
+    const applyingRef = useRef(false);
```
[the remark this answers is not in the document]
tomasfratrik: This also seems like a good refactor if im not missing anything, so I will implement this

src/actions/payload-dnf-actions.js
```diff
@@ -76,7 +84,53 @@ export const getPayloadGroupsAction = (environment) => {
...
-            type: "SET_PAYLOAD_GROUPS"
+            type: "SET_PAYLOAD_GROUPS",
         });
     };
 };
+
+/**
+ * Reload environments and re-validate the current selection after the source changes.
+ */
+export const refreshPayloadSoftwareSelectionAction = () => {
```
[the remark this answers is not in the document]
tomasfratrik: I dont think I will do this

tomasfratrik:
I have reviewed and implemented some of the bots comments, and in latest commit I added the screen hidding

tomasfratrik:
Note: installer doesn't acknowledge new installation source at the end screen.

For example when I was installing f45 rawhide and using f43 source pkgs (so essentially you will have f43 after reboot), at the end of installer you are shown `"To begin using F45 rawhide, reboot your system"`. I am not sure, but I think this is because it takes version from installation media `/etc/os-release` but doesn't acnowledge the fact that that,  doesn't have to be the final system. 

(Maybe not relevant but I think that the `os-release` is determined by the `fedora-release` pkg)

I believe just saying "reboot to your new system" without version, is better then saying it with wrong version.
But obviously in perfect world we would want it to say the same release as will be in the rebooted system. (even somehow if you mix pkgs)

Another easier solution would be to only show release at the end if user didn't use any custom installation source

src/components/review/ReviewConfiguration.jsx
```diff
@@ -154,6 +159,31 @@ export const ReviewConfiguration = ({ autoProceedBlockedRef, automatedInstall, p
 
     const installationScenarioDescription = <StorageScenarioReviewDescription />;
 
+    const sourceDescription = useMemo(() => {
```
KKoukiou: please don't add here inline all these. See how we do that for other screens ex:

import { AccountsReviewDescription } from "../users/index.js";

src/components/source/InstallationSource.jsx
```diff
@@ -0,0 +1,331 @@
...
+    return (
+        <Form id={SCREEN_ID}>
+            {isApplying && (
+                <Alert
+                  isInline
+                  title={_("Setting up installation source")}
+                  variant="info"
+                />
+            )}
+            {error && (
```
KKoukiou: Use const { setStepNotification } = useContext(PageContext) ?? {};

This will add the error in the correct place.

src/components/source/InstallationSource.jsx
```diff
@@ -0,0 +1,331 @@
...
+        [applyCurrentSource, isDirty]
+    );
+    useWizardFooter(footer);
+
+    return (
+        <Form id={SCREEN_ID}>
+            {isApplying && (
+                <Alert
+                  isInline
+                  title={_("Setting up installation source")}
```
KKoukiou: So I was testing this, and this confused me - I did not realize that the was an in progress operation. I would suggest to extend src/components/AnacondaWizardFooter.jsx to support isLoading property in the "next" button.

KKoukiou: And while the check is performed disable the form.

src/actions/payload-dnf-actions.js
KKoukiou, line 95: I think we have a small regression in the behavior compared to before. 

When entering the Sofrware selection the additional software is not correctly prefilled. 

Image

after one clicked on the encvironment selction is get's selected: 

Image

KKoukiou (commented):
We don't have to support all sources at once in the configuration, but we need to be able to allow the user that has configured through KS some other source like cdrom to proceed through this new page withtu changing their KS selections. That means we need to show at least Read Only the user selection and respect that.

src/components/review/ReviewConfiguration.jsx
tomasfratrik: definitely true, also I noticed some left over code from when I was reducing it.

src/components/source/InstallationSource.jsx
tomasfratrik: This is much better yes

tomasfratrik:
Another thing I noticed: when you input some url, then go back to closest mirror, the url persists there (just visuall impact), so I will fix this.  Also upon entering the sw selection, there is brief loading hint again, to refresh sw, this is also done in the install source, so maybe I can remove this, and if not I will have to change this to some better loader

Another bug: while source install is loading you can edit the input box (should be greyed out)

src/actions/payload-dnf-actions.js
tomasfratrik: For some reason Im unable to recreate this.. using this PRs exact code, the same testvm server edition and I always have the additional software present immediately right after clicking next from install source (with server preselected as you do), and thats for both closest mirror and some custom url

tomasfratrik: I finally have  have recreated some version of it -> you put there invalid link so it errors, than you go BACK, and then next (you are autmatically on closest mirror) and next to sw selection and now you have nothing there to select, not even environment,.. (the frontend selection of closest mirror lets you pass to sw selection page, but the source is still corrupted) so then when you go BACK to source install and input some valid url, and go NEXT to sw selection, then I see page like you did where I have to click on that env to have additional sw. 
This didin't work for me without going BACK to date & time.

tomasfratrik: So after some investigation, this occured because the `isDirty` value is reset to false when going to 'back' page because it matches the `initialUiState`. The solution could be a force install again of closest mirror, but that isn't good. I will introduce a new redux flag that remembers whether the apply failed so you can't continue. I will put this fix into separate commit for better visibility, can squash later.

src/components/source/InstallationSource.jsx
tomasfratrik: fixed

tomasfratrik:
> Another thing I noticed: when you input some url, then go back to closest mirror, the url persists there (just visuall impact), so I will fix this. Also upon entering the sw selection, there is brief loading hint again, to refresh sw, this is also done in the install source, so maybe I can remove this, and if not I will have to change this to some better loader
> 
> Another bug: while source install is loading you can edit the input box (should be greyed out)

So I fixed both Input URL bugs... and with the hint in the sw selection, It is just defensive duplicate reload, which I don't think is needed and was more of implemented when the screen was erroring on sw selection but that was due to other bug IIRC

tomasfratrik:
Note: the spinner on button is white and it is spinning on disabled button which is grey, and I think the visibility isn't the best

src/components/source/SourceReviewDescription.jsx
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (C) 2026 Red Hat, Inc.
```
bruno-fs, line 2: This is a very silly nitpick but I found interesting some of the copyright notices are using current year, while others (in new files) use 2025 🤔. Out of the 9 new files, 7 are using a copyright notice from 2025.

I was thinking some check for copyright notices might help spot these things, but I'm almost certain cockpit has one enabled in their test infra.

tomasfratrik: That is interesting, nothing a one `sed` thru whole project can't do :smile:

tomasfratrik: should be fixed

tomasfratrik:
I tested the ks support for the http/s, metalink, mirrorlist, but will have to to ks support for other installs (as read only)

tomasfratrik:
The installation source screen only knows a closest mirror and plain http/s
URLs. Sources it can not express — NFS, FTP, hard drive, local media, are now
handled like this:

* the install source screen is dropped from the wizard 
* the source is shown read-only in the Review screen, with a hint that it can
  be changed in the kickstart file
* the text comes from the backend `Source.Description` property, so the Web UI
  does not parse every source type itself

For http/s URLs nothing changes, they stay editable in the installation source
screen. The options the screen has no field for are kept when the URL is
re-applied, and shown read-only next to it so it is clear they are still in
effect: proxy, SSL CA certificate, SSL client certificate and disabled SSL
verification (`--noverifyssl`). Passwords in URLs (proxy, FTP) are masked as
`user:****@host`, in the review screen.

Screenshots of how each case looks are in the ticket [INSTALLER-3621](https://redhat.atlassian.net/browse/INSTALLER-3621), together with the kickstart files I tested with and the host setup they need (serving the tree
over FTP/HTTP, the NFS export, building the disk image for `harddrive`).

The second commit is test harness only: `--add-disk` can now be given more than
once and takes either a size or a path to a prepared image, which is what the
`harddrive` kickstart needs. It is unrelated to the source work, so it should
probably go in as a separate PR, happy to split it out.

PR #1385, opened 2026-07-24
storage: Skip partitioning reset for kickstart scenarios

KKoukiou:
The usePartitioningReset hook was resetting the partitioning when entering the Installation Method page for the kickstart scenario. The condition excluded "use-configured-storage" (cockpit storage) but not "use-configured-storage-kickstart", causing an unnecessary reset/re-apply cycle.

The pixel test changes are desired - there was some race before causing the swap to dissapear. Now it's there as the test specifies; https://github.com/rhinstaller/anaconda-webui/blob/main/test/kickstarts/TestKickstartAutomated-testBasic.ks#L6


rvykydal (approved):
Looks good to me.

PR #1393, opened 2026-07-28
Add quick-updates.img target for faster dev cycles

bruno-fs:
Build updates.img directly from sources, bypassing RPM packaging and the build VM. Reduces build time from ~60s to ~3s*. Optionally merges anaconda backend changes via ANACONDA_DIR - which also saves time given building the backend RPM is also a time consuming process.

Also reduce webui_testvm.py readiness polling interval from 10s to 2s, reducing VM startup time from ~60s to ~45s*.

* Measured on the author's machine.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/HACKING.rst
+++ b/HACKING.rst
@@ -33,7 +33,7 @@ To prepare Anaconda Web UI sources, you need to run this command::
     make dist
 
 The easiest way to test changes you make is to set up a test VM.
-You can find intructions for preparing a test VM at ``test/README.rst``.
+You can find instructions for preparing a test VM at ``test/README.rst``.
 
 Running eslint
 --------------
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,10 @@ create-updates.img: bots
 	-rm *updates.img
 	make $(UPDATES_IMG)
 
+# Download dependency RPMs (cockpit, anaconda-core, udisks) via container.
+fetch-rpms:
+	test/fetch-rpms
+
 test/reference: test/common
 	test/common/pixel-tests pull
 
--- a/Makefile
+++ b/Makefile
@@ -221,6 +221,26 @@ create-updates.img: bots
 fetch-rpms:
 	test/fetch-rpms
 
+tmp/rpms:
+	$(MAKE) fetch-rpms
+
+# Fast updates.img for dev: builds JS + installs files directly, no RPM/VM.
+# Set ANACONDA_DIR=~/src/anaconda to also include backend changes.
+# Fetches dependency RPMs on first run, then reuses them.
+quick-updates.img: $(DIST_TEST) tmp/rpms
+	rm -rf updates updates.img
+	$(MAKE) install DESTDIR=$$(pwd)/updates
+	cd updates && for rpm in ../tmp/rpms/*.rpm; do rpm2cpio "$$rpm" | cpio -idmu 2>/dev/null; done
+	if [ -n "$${ANACONDA_DIR}" ]; then \
+		tag=$$(cd "$${ANACONDA_DIR}" && git describe --tags --abbrev=0 2>/dev/null) && \
+		(cd "$${ANACONDA_DIR}" && ./scripts/makeupdates -t "$$tag") && \
+		(cd updates && gzip -dc "$${ANACONDA_DIR}/updates.img" | cpio -idmu 2>/dev/null); \
+	fi
+	cd updates && find . | cpio -c -o 2>/dev/null | gzip -9 > ../updates.img
+	rm -rf updates
+	cp updates.img updates-$$(echo $(TEST_OS) | sed 's/-boot//').img
+	@ls -lh updates.img updates-$$(echo $(TEST_OS) | sed 's/-boot//').img
+
 test/reference: test/common
 	test/common/pixel-tests pull
 
--- a/test/machine_install.py
+++ b/test/machine_install.py
@@ -223,12 +223,12 @@ class VirtInstallMachine(VirtMachine):
             if not self.is_live():
                 Machine.wait_boot(self, timeout_sec=300)
 
-                for _ in range(30):
+                for _ in range(150):
                     try:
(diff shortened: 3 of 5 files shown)


HACKING.rst
```diff
@@ -33,7 +33,26 @@ To prepare Anaconda Web UI sources, you need to run this command::
     make dist
 
 The easiest way to test changes you make is to set up a test VM.
-You can find intructions for preparing a test VM at ``test/README.rst``.
+You can find instructions for preparing a test VM at ``test/README.rst``.
+
+Quick dev iteration with updates.img
```
KKoukiou: I would prefer to keep it together with the plain `make create-updates.img` docs

That's in test/README.rst now. 

Also there you can find instructions how to add extra RPMs.

```                                                                                                                                                             
 23 You can optionally add extra RPMs to the updates.img by setting the ``TEST_RPMS`` variable,        
 24 with a comma-seperated list of RPMs to add. For example::                                          
 25                                                                                                                                                             
 26     make TEST_RPMS="/tmp/anaconda-core-123.rpm, /tmp/anaconda-core-debuginfo-123.rpm" create-updates.img
```

So the reason we were building the RPMs in a VM is that we make sure they are built on the same fedora version like the target installation. This is especialy important for python files as the path changes with different python versions.

test/machine_install.py
```diff
@@ -223,12 +223,12 @@ def start(self):
             if not self.is_live():
                 Machine.wait_boot(self, timeout_sec=300)
 
-                for _ in range(30):
+                for _ in range(150):
```
KKoukiou, line 226: The polling change should be a separate commit - it's a nice improvement but is unrelated to the  quick-updates.img target

Makefile
```diff
@@ -217,6 +217,20 @@ create-updates.img: bots
 	-rm *updates.img
 	make $(UPDATES_IMG)
 
+# Fast updates.img for dev: builds JS + installs files directly, no RPM/VM.
+# Set ANACONDA_DIR=~/src/anaconda to also include backend changes.
+quick-updates.img: $(DIST_TEST)
+	rm -rf updates updates.img
+	$(MAKE) install DESTDIR=$$(pwd)/updates
+	if [ -n "$${ANACONDA_DIR}" ]; then \
+		tag=$$(cd "$${ANACONDA_DIR}" && git describe --tags --abbrev=0 2>/dev/null || echo "HEAD~1") && \
```
KKoukiou: There are always tags so the fallback part is not needed.

Makefile
```diff
@@ -217,6 +217,20 @@ create-updates.img: bots
 	-rm *updates.img
 	make $(UPDATES_IMG)
 
+# Fast updates.img for dev: builds JS + installs files directly, no RPM/VM.
```
KKoukiou, line 227: So for front end changes - the faster way to iterate is to run `make rsync` against the running VM. We don't need to reboot or anything.

KKoukiou (commented):
I tried this and it does not work.

The reason is does not work is that there are missing dependencies that are required to run the WebUI installer. (cockpit for sure)

``` 
`Jul 29 08:11:29 localhost.localdomain systemd[1]: Started webui-cockpit-ws.service - Cockpit Web Service for Anaconda Installer.
Jul 29 08:11:29 localhost.localdomain cockpit-coproc-wrapper.sh[3485]: /usr/libexec/anaconda/cockpit-coproc-wrapper.sh: line 17: /usr/libexec/cockpit-ws: No such file or directory
Jul 29 08:11:29 localhost.localdomain systemd[1]: webui-cockpit-ws.service: Main process exited, code=exited, status=127/n/a
Jul 29 08:11:29 localhost.localdomain systemd[1]: webui-cockpit-ws.service: Failed with result 'exit-code'.
```

The VM when spawned with ./test/webui-testvm uses updates-fedora-rawhide.img  - not updates.img

I think you built the updates.img but kept using the old updates-fedora-rawhide.img thinking that it got updated.

HACKING.rst
bruno-fs: Mmmm  so can can just use a lightweight fedora container for that

KKoukiou: yes the rpms are build within `anaconda-rpm-build` containers in anaconda repo

bruno-fs: while that works, it's a slower process than what I'm proposing here - especially if you are iterating on changes both on backend and frontend.

Makefile
bruno-fs: I'm aware :)

the goal is to have a faster way to iterate on both backend and ui changes at the same time

HACKING.rst
KKoukiou: @bruno-fs sure but the proposal does not really work because this solution is not pulling in dependencies. 

I think what you can do it teach the script to only take into an existing updates.img the anaconda changes while keeping the rest from previous build. That will work and will be fast.

M4rtinK: If (part) of the issue is missing packages needed on the boot.iso - I say lets add them there. It won't make the image significantly larger & prepares it for the eventual switch. We might even find some issues already during the F45 cycle even without actually shipping the Web UI.

tomasfratrik:
lgtm, but I would maybe want 1 QOL change, when you abort (can be some network failure, maybe corp down ?) during the quick-updates the fetch-rpms gets interrupted and the deps list will be just partial, and on next quick-updates sees the tmp/rpms folder and skips the fetch-rpms, but then the image builds with missing deps and the person might not know this... maybe this could just be resolved with fetch-rpms creating some "done" stamp file like .fetched at the end so we know all was fetched, but there are multiple ways to do this ..

tomasfratrik:
lgtm, but I would maybe want 1 QOL change, when you abort (can be some network failure, maybe corp down ?) during the quick-updates the fetch-rpms gets interrupted and the deps list will be just partial, and on next quick-updates sees the tmp/rpms folder and skips the fetch-rpms, but then the image builds with missing deps and the person might not know this... maybe this could just be resolved with fetch-rpms creating some "done" stamp file like .fetched at the end so we know all was fetched, but there are multiple ways to do this ..

PR #1394, opened 2026-07-29
Add Bugzilla duplicate detection to the error reporting flow

adamkankovsky:
Search for similar open bugs before submitting a new report so users can attach information to an existing issue instead of creating duplicates.
image

### DIFF
--- a/test/helpers/bugzilla_mock.py
+++ b/test/helpers/bugzilla_mock.py
@@ -1,0 +1,73 @@
+# Copyright (C) 2025 Red Hat, Inc.
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""Helpers for mocking the python-bugzilla package in integration tests."""
+
+import json
+from pathlib import Path
+
+MOCK_BUGS_PATH = "/tmp/anaconda-bugzilla-mock-bugs.json"
+
+MOCK_SCRIPTS_DIR = Path(__file__).parent.parent / "mock_bugzilla_scripts"
+
+
+def _python_cmd(machine):
+    return machine.execute(
+        "command -v /usr/libexec/platform-python || command -v python3"
+    ).strip()
+
+
+def _site_packages(machine):
+    py = _python_cmd(machine)
+    return machine.execute(
+        f"{py} -c 'import sysconfig; print(sysconfig.get_paths()[\"purelib\"])'"
+    ).strip()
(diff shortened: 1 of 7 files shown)


rvykydal (approved):
Looks good to me.

test/helpers/bugzilla_mock.py
```diff
@@ -0,0 +1,116 @@
...
+        return [Bug(item["id"], item["summary"], item["status"]) for item in data]
+
+    def createbug(self, **kwargs):
+        class NewBug:
+            id = 999999
+        return NewBug()
+
+    def attachfile(self, *args, **kwargs):
+        return 1
+'''
```
bruno-fs: while this works, I'd prefer to see this in an actual file instead of a string. we could easily load that later with pathlib's Path object like the following `Path("path/to/file").read_text()`

test/helpers/bugzilla_mock.py
```diff
@@ -0,0 +1,116 @@
...
+SITE="{site}"
+if [ -e "$SITE/bugzilla" ] && [ ! -e "$SITE/bugzilla.anaconda-test-bak" ]; then
+    mv "$SITE/bugzilla" "$SITE/bugzilla.anaconda-test-bak"
+fi
+if [ -e "$SITE/bugzilla.py" ] && [ ! -e "$SITE/bugzilla.py.anaconda-test-bak" ]; then
+    mv "$SITE/bugzilla.py" "$SITE/bugzilla.py.anaconda-test-bak"
+fi
+mkdir -p "$SITE/bugzilla"
+""")
+    machine.write(f"{site}/bugzilla/__init__.py", MOCK_BUGZILLA_INIT)
```
bruno-fs:
```suggestion
    MOCK_BUGZILLA_INIT = Path("path/to/bzmock.py").read_text()
    machine.write(f"{site}/bugzilla/__init__.py", MOCK_BUGZILLA_INIT)
```

bruno-fs (commented):
The feature is very much needed and I love that you even put together nice mocks to be able to test it! kudos for that

The only thing I would change is all the files that are stored as string constants. We could instead simply have the actual files as some test utilities (maybe under `test/mock_bugzilla_scripts/`) and just copy them to the vm.

bruno-fs (approved):

PR #1404, opened 2026-08-05
 feat: track installation status and pending errors on reconnection 

bruno-fs:
Add Boss context to track installation status and pending errors, allowing the progress screen to recover state when a remote user reconnects. Rework InstallationProgress to branch on backend status instead of always starting a new installation.

Add reconnection tests for non-critical and fatal error cases.

Resolves: INSTALLER-4815
Related: INSTALLER-4824
Assisted-by: Claude Opus 4.6

### DIFF
--- a/src/components/app.jsx
+++ b/src/components/app.jsx
@@ -34,7 +34,7 @@ export const ApplicationLoading = () => (
     </PageSection>
 );
 
-export const Application = ({ conf, dispatch, isFetching, onCritFail, osRelease, reportLinkURL, setShowStorage, showStorage }) => {
+export const Application = ({ conf, dispatch, installationStatus, isFetching, onCritFail, osRelease, reportLinkURL, setShowStorage, showStorage }) => {
     const [storeInitialized, setStoreInitialized] = useState(false);
     const [currentStepId, setCurrentStepId] = useState();
     const address = useAddress(onCritFail);
@@ -75,7 +75,7 @@ export const Application = ({ conf, dispatch, isFetching, onCritFail, osRelease,
     }, [address, automatedInstall, conf, dispatch, onCritFail]);
 
     // Postpone rendering anything until we read the dbus address and the default configuration
-    if (!address || !storeInitialized) {
+    if (!address || !storeInitialized || !installationStatus) {
         debug("Loading initial data...");
         return <ApplicationLoading />;
     }
@@ -210,6 +210,7 @@ export const ApplicationWithErrorBoundary = () => {
                     <Application
                       conf={conf}
                       dispatch={dispatch}
+                      installationStatus={state.boss.installationStatus}
                       isFetching={state.misc.isFetching}
                       onCritFail={onCritFail}
                       osRelease={osRelease}
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -55,6 +55,8 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
     useAutoReboot(status, automatedInstall);
 
     useEffect(() => {
+        const failureCtx = { context: _("Installation of the system failed") };
+
         const connectToTask = (taskPath, shouldStart) => {
             const taskProxy = new BossClient().client.proxy(
                 "org.fedoraproject.Anaconda.Task",
@@ -116,9 +118,7 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
             taskProxy.wait(() => {
                 addEventListeners();
                 if (shouldStart) {
-                    taskProxy.Start().catch(onCritFail({
-                        context: _("Installation of the system failed"),
-                    }));
+                    taskProxy.Start().catch(onCritFail(failureCtx));
                 } else {
                     getSteps({ task: taskPath })
                             .then(
@@ -137,14 +137,10 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                         installWithTasks()
                                 .then(
                                     tasks => connectToTask(tasks[0], true),
-                                    onCritFail({
-                                        context: _("Installation of the system failed"),
-                                    })
+                                    onCritFail(failureCtx)
                                 );
                     }
-                }, onCritFail({
-                    context: _("Installation of the system failed"),
-                }));
+                }, onCritFail(failureCtx));
     }, [onCritFail]);
 
     const submitErrorDecision = (shouldContinue) => {
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -32,6 +32,7 @@ const _ = cockpit.gettext;
 const N_ = cockpit.noop;
 const SCREEN_ID = "anaconda-screen-progress";
 const DETAIL_TYPE_YESNO = "yesno";
+const PROGRESS_STEPS_DONE = 4;
 
 const progressStepsMap = {
     BOOTLOADER_INSTALLATION: 2,
@@ -109,7 +110,7 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                 });
                 taskProxy.addEventListener("Succeeded", () => {
                     setStatus("success");
-                    setCurrentProgressStep(4);
+                    setCurrentProgressStep(PROGRESS_STEPS_DONE);
                 });
             };
             taskProxy.wait(() => {
@@ -211,11 +212,11 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
               paragraph={
                   <Flex direction={{ default: "column" }}>
                       <Content component="p">
-                          {currentProgressStep < 4
+                          {currentProgressStep < PROGRESS_STEPS_DONE
                               ? progressSteps[currentProgressStep].description
                               : cockpit.format(_("To begin using $0, reboot your system."), osRelease.PRETTY_NAME)}
                       </Content>
-                      {currentProgressStep < 4 && (
+                      {currentProgressStep < PROGRESS_STEPS_DONE && (
                           <>
                               <FlexItem spacer={{ default: "spacerXl" }} />
                               <ProgressStepper isCenterAligned>
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -101,16 +101,7 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                     });
                 });
                 categoryProxy.addEventListener("ErrorRaised", (_, message, detailType) => {
-                    if (detailType === DETAIL_TYPE_YESNO) {
-                        setErrorDialogData({
-                            categoryProxy,
-                            message,
-                        });
-                    } else {
-                        setStatus("danger");
-                        categoryProxy.RespondToError(false);
-                        onCritFail()({ message });
-                    }
+                    handleError(message, detailType, categoryProxy);
                 });
                 taskProxy.addEventListener("Succeeded", () => {
                     setStatus("success");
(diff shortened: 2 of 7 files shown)


src/components/installation/InstallationProgress.jsx
```diff
@@ -124,27 +128,41 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
...
-                    context: _("Installation of the system failed"),
-                }));
-    }, [onCritFail]);
+        if (installationStatus === INSTALLATION_STATUS.SUCCEEDED) {
+            setStatus("success");
+            setCurrentProgressStep(PROGRESS_STEPS_DONE);
+            setSteps([]);
+        } else if (installationStatus === INSTALLATION_STATUS.FAILED) {
+            onCritFail()({ message: pendingError.message });
+        } else if (installationStatus === INSTALLATION_STATUS.RUNNING) {
```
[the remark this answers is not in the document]
bruno-fs: being undefined is not an issue given the initial state this PR sets for boss
```js
 export const bossInitialState = {
    installationStatus: null,
    pendingError: { message: "", type: "" },
};
```

but we might wanto to use a fallbackmessage

src/components/installation/InstallationProgress.jsx
```diff
@@ -110,7 +114,7 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                 });
                 taskProxy.addEventListener("Succeeded", () => {
                     setStatus("success");
-                    setCurrentProgressStep(4);
+                    setCurrentProgressStep(PROGRESS_STEPS_DONE);
```
KKoukiou, line 108: Belongs to the previous commit

src/components/installation/InstallationProgress.jsx
```diff
@@ -125,27 +129,48 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
...
+        } else if (installationStatus === INSTALLATION_STATUS.FAILED) {
+            onCritFail()({ message: pendingError.message });
+        } else if (installationStatus === INSTALLATION_STATUS.RUNNING) {
+            getActiveInstallationTask().then(activeTask => {
+                if (activeTask) {
+                    connectToTask(activeTask, false);
+                } else {
+                    // this should be impossible, but will guard against (and log) just in case...
+                    debug("Installation status is RUNNING but no active task found, starting installation");
+                    installWithTasks().then(
```
KKoukiou: I guess since this is impossible we should not try to start the installation again.. I would drop this code apart from the debug log.

src/components/installation/InstallationProgress.jsx
```diff
@@ -125,27 +129,48 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
...
-                }));
-    }, [onCritFail]);
+        if (installationStatus === INSTALLATION_STATUS.SUCCEEDED) {
+            setStatus("success");
+            setCurrentProgressStep(PROGRESS_STEPS_DONE);
+            setSteps([]);
+        } else if (installationStatus === INSTALLATION_STATUS.FAILED) {
+            onCritFail()({ message: pendingError.message });
+        } else if (installationStatus === INSTALLATION_STATUS.RUNNING) {
+            getActiveInstallationTask().then(activeTask => {
```
KKoukiou: You can also use if you like: const activeTask = awaita getActiveInstallationTask, it will make the code nicer.

src/components/app.jsx
```diff
@@ -69,13 +70,17 @@ export const Application = ({ conf, dispatch, isFetching, onCritFail, osRelease,
         document.addEventListener("click", allowExternalNavigation);
 
         new BossClient(address, dispatch).init({ automatedInstall, conf })
+                .then(() => Promise.all([
```
KKoukiou: These belong to the init method of the BossClient.

rvykydal (approved):
Looks good to me.
I wonder if it makes sense to add reload() check (test) also for installation successfully completed.

bruno-fs:
> I wonder if it makes sense to add reload() check (test) also for installation successfully completed.

There's a bug (INSTALLER-4825) preventing that from working properly. I shall fix it soon, I have a patch for that =D

src/components/installation/InstallationProgress.jsx
```diff
@@ -41,6 +40,8 @@ const progressStepsMap = {
     SYSTEM_CONFIGURATION: 3,
 };
 
+const PROGRESS_STEPS_DONE = Math.max(...Object.values(progressStepsMap)) + 1;
```
M4rtinK: Seem a bit complicated, why not just assign it a number like the other steps ? I don't think we expect those to be very dynamic.

src/components/installation/InstallationProgress.jsx
```diff
@@ -125,27 +129,48 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
...
-                }, onCritFail({
-                    context: _("Installation of the system failed"),
-                }));
-    }, [onCritFail]);
+        if (installationStatus === INSTALLATION_STATUS.SUCCEEDED) {
+            setStatus("success");
+            setCurrentProgressStep(PROGRESS_STEPS_DONE);
+            setSteps([]);
+        } else if (installationStatus === INSTALLATION_STATUS.FAILED) {
+            onCritFail()({ message: pendingError.message });
```
M4rtinK: Does this also somehow handle also non fatal errors ? E.g. those where user has the additional option of continuing the installation instead of the always present option to abort.

src/reducer.js
```diff
@@ -42,6 +42,14 @@ export const localizationInitialState = {
     xlayouts: undefined,
 };
 
+export const INSTALLATION_STATUS = { FAILED: 4, NOT_STARTED: 1, RUNNING: 2, SUCCEEDED: 3 };
```
M4rtinK: Wasn't there already a listing of states defined in the first commit, or is this different ?

M4rtinK (approved):
Looking good - just adding a few questions/suggestions. :)

src/components/installation/InstallationProgress.jsx
bruno-fs: non fatal errors are handled in a distinct code path. they don't have a specific status (they occur during the RUNNING phase and are detected due to the presence of the pending errors)

src/reducer.js
bruno-fs: no. the first commit is tangent to the installation progress. 

maybe using a int enum for tracking this particular state was a mistake. using actual strings might make this clearer. what do you think?

src/components/installation/InstallationProgress.jsx
```diff
@@ -129,45 +120,56 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
...
-                            setErrorDialogData({
-                                categoryProxy,
-                                message: pendingError.message,
-                            });
-                        } else {
-                            setStatus("danger");
-                            onCritFail()({ message: pendingError.message });
-                        }
-                    }
+                    handleError(pendingError.message, pendingError.type, categoryProxy);
```
bruno-fs, line 121: here is where we are handling non critical errors

src/components/installation/InstallationProgress.jsx
bruno-fs: I commented there to make it easier to link back here

https://github.com/rhinstaller/anaconda-webui/pull/1404/commits/1e10d46c0366073d4345d519a3e1f6e1e41a8d5b#r3752711329

src/components/installation/InstallationProgress.jsx
bruno-fs: yeah, maybe it could be achieved in a simpler way. my pythonista 🐍 brain was expecting something like `max(progress_step_map.values()) +1`.

I initially set it to 4, and then thought "huh, but if we update these steps?"

src/reducer.js
bruno-fs:
```suggestion
export const INSTALLATION_STATUS = { FAILED: "FAILED", NOT_STARTED: "NOT_STARTED", RUNNING: "RUNNING", SUCCEEDED: "SUCEEDED" };
```
If we change to strings would this be more clear? I will need the backend too to match this

src/components/installation/InstallationProgress.jsx
M4rtinK: Well, this is all in Javascript, right ? E.g. not passing the status codes via RPC from the Python backend to the Javascript frontend. So it should always be a constant & its not really relevant what the value is, just that it is distinct and always maps to the same constant. We also don't need support long term persistence.

So if we add another state, we just add another constant with a distinct value.

src/components/installation/InstallationProgress.jsx
M4rtinK: Sounds good then - I guess that we make sure to show even non fatal errors on reconnection, right ? :)

src/reducer.js
M4rtinK: Yep, I like this much more - as I've mentioned in my previous comment, we really should always use the constants across the whole Web UI codebase & never do comparisons against anything else than the constants.

But having the constant value actually by a descriptive string also helps - if things crash or the value ends up somewhere wrong, it will be much easier to identify what went wrong. :)

src/components/installation/InstallationProgress.jsx
M4rtinK: ACK, thanks. :)

src/components/app.jsx
bruno-fs: I was afraid of adding circular imports there (boss-actions import `getInstallationStatus`  from `BossClient`, then `BossClient` imports `getInstallationStatusAction` from boss-actions back?). It will work if I get rid of boss-actions.js... WDYT?

src/components/installation/InstallationProgress.jsx
bruno-fs: k, will set it to 4. if it ever needs updates it won't go unnoticed anyways...

src/components/app.jsx
KKoukiou: Fine to leave this as is - sorry

bruno-fs: no problem :)

src/components/installation/InstallationProgress.jsx
bruno-fs: yes, this is the bit that's doing that. There's another part of the code that is handling that based on the event, I think. Futhermore, this PR is adding a page reload to that causes a non critical error and making sure the error is visible :)

src/components/installation/InstallationProgress.jsx
bruno-fs: mmmm if it wasn't for the error handling I would agree, but including error handling it looks almost the same thing to me, but doesn't follow the pattern on the rest of the code IMO
```js
try {
          const activeTask = await getActiveInstallationTask();
          if (activeTask) {
              connectToTask(activeTask, false);
          } else {
              debug("Installation status is RUNNING but no active task found, starting installation");
              startNewInstallation();
          }
      } catch (error) {
          onCritFail({ context: _("Installation of the system failed") })(error);
      }
```
unless we also refactor `getActiveInstallationTask` to autocatch errors and run `onCritFail`, then, yeah, it would be a lot nicer. just

```
const activeTask = await getActiveInstallationTask();
```
mmmm maybe I should do this refactor too 🤔 

EDIT: ugh, no that wont work and I guess we also need to catch errors raised by connectToTask - that's the risky one. I will keep it as is for now because I couldn't figure out a way to make it look nicer 😢 

EDIT2: @KKoukiou told me in a DM we are moving away from the other pattern. In this case I will change things to use await.

src/reducer.js
bruno-fs: https://github.com/rhinstaller/anaconda/pull/7230

src/components/installation/InstallationProgress.jsx
bruno-fs: should we throw an error too? because I don't see a way for the installation to continue if this happens and users would get stuck here.

src/components/installation/InstallationProgress.jsx
bruno-fs: done! thanks for catching that :)

src/components/installation/InstallationProgress.jsx
bruno-fs: any thoughts, @M4rtinK ?

bruno-fs: Ok, with a fresh perspective I made my mind: I think this needs to stay because this is a harmless call. If it is redundant (and it should be), the backend is already guarded against multiple calls https://github.com/rhinstaller/anaconda/blob/cafe81f3058cc31c03cd734ffc1d4469c572d669/pyanaconda/modules/boss/boss.py#L191-L193

If for whatever reason this branch is being reached we would be able to spot this message on logs.

Alternatively, we could throw an error to fail loudbly.

bruno-fs: we decided to go with an audible failure!

PR #1408, opened 2026-08-10
Test incomplete kickstart timezone

adamkankovsky:


test/kickstarts/TestKickstartAutomatedIncompleteTimezone-testBasic.ks
```diff
@@ -0,0 +1,23 @@
+bootloader --timeout=1
```
KKoukiou: test/kickstarts/TestKickstartAutomatedIncomplete-testBasic.ks

This covers this scenario. Can't you reuse that test?

src/apis/timezone.js
```diff
@@ -104,8 +104,11 @@ export const getKickstarted = () => {
  * @param {string} params.timezone - The timezone to set (e.g. 'Europe/Prague')
  * @returns {Promise<void>}
  */
-export const setTimezone = ({ timezone }) => {
-    return callClient("SetTimezoneWithPriority", [timezone, 90]);
+export const setTimezone = async ({ timezone }) => {
+    await callClient("SetTimezoneWithPriority", [timezone, 90]);
+    // Mark timezone as user-configured so incomplete kickstarts (no timezone
+    // command) can proceed after the user sets one in the Date & Time UI.
+    TimezoneClient.instance?.dispatch?.(setTimezoneAction({ timezone, userConfigured: true }));
```
KKoukiou: We should listen to Timezone PropertyChanges signal and then emit an action asynchronously - not do it here.

KKoukiou (changes requested):

test/kickstarts/TestKickstartAutomatedIncompleteTimezone-testBasic.ks
adamkankovsky: Unfortunately, that won't work it's a different Kickstart and, more importantly, a different workflow.

src/components/datetime/usePageComplete.js
```diff
@@ -22,7 +23,8 @@ export const usePageComplete = ({ automatedInstall, isHidden } = {}) => {
         return true;
     }
     const kickstarted = tz?.kickstarted;
-    if (automatedInstall && !kickstarted) {
+    const userConfigured = tz?.userConfigured;
+    if (automatedInstall && !kickstarted && !userConfigured) {
```
KKoukiou, line 27: Can't we just check that the timezone is set? (aka timezone is not None)?

test/check-kickstart-automated
```diff
@@ -276,43 +276,64 @@ class TestKickstartAutomatedIncomplete(VirtInstallMachineCase):
...
 
-        # Wait for completeness of accounts
-        # Note: also makes sure that the button state stabilizes for the next check
         b.wait_in_text(f"#{i.steps.REVIEW}-target-system-account",
                        "The following user will be created")
+        r.check_timezone("Europe/Prague")
         i.check_next_disabled(disabled=False)
 
+        i.begin_installation(needs_confirmation=False, button_text="Install")
+        p.wait_done()
```
KKoukiou: I don't see the reason that the installation is performed, since we dont check anything post reboot. Its enough to check that the installation can actually start (button is not disabled) and keep the tests non-destructive.

test/check-kickstart-automated
```diff
@@ -280,37 +280,50 @@ class TestKickstartAutomatedIncomplete(VirtInstallMachineCase):
...
         i = Installer(b, m, scenario="use-configured-storage-kickstart")
+        r = Review(b, m)
+        dt = DateAndTime(b, m)
         u = Users(b, m)
 
         b.open("/cockpit/@localhost/anaconda-webui/index.html")
 
-        # Wait for warning notification about incompleteness of accounts
-        # Note: also makes sure that the button state stabilizes for the next check
         b.wait_in_text(f"#{i.steps.REVIEW}-incomplete-configuration",
```
KKoukiou, line 301: We can / should check that both accounts and timezone are marked with warning here.

KKoukiou (approved):

PR #1409, opened 2026-08-10
Require authentication for remote installation

bruno-fs:
This work was built on top of the proof of concept #1274 and rhinstaller/anaconda#7228.

Now `inst.webui.remote` requires one additional argument: either `inst.webui.remote.pin=<somevalue>`, which is the intended/recommended way for end users to use it, or `inst.webui.remote.noauth`, which mimicks current behavior. This last option was added as convenience to not disrupt tests.

## How to test

The new test requires the backend to work, so for now it can only be tested with something like
```
TEST_SCENARIO=anaconda-pr-7228 make create-updates.img
toolbox run -c anaconda-webui bash -c 'TEST_OS=fedora-rawhide-boot TEST_AUDIT_NO_SELINUX=1 test/check-remote-auth'
```
Then you can run the vm with webui_testvm CLI
```
toolbox run -c anaconda-webui python test/webui_testvm.py fedora-rawhide-boot --remote-pin=<SOMEPIN>
```

## Screenshots

When logging in you should see this
image

passing an incorrect PIN should trigger the following and clicking "login" should display the following
image

finally, clicking the 👁️ reveals the PIN value
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,11 +128,17 @@ install: $(DIST_TEST) po/LINGUAS
 	cp browser-ext $(DESTDIR)/usr/libexec/anaconda
 	cp gnome-control-center-ext $(DESTDIR)/usr/libexec/anaconda
 	cp src/scripts/cockpit-coproc-wrapper.sh $(DESTDIR)/usr/libexec/anaconda/
+	cp src/scripts/cockpit-pin-auth $(DESTDIR)/usr/libexec/anaconda/
 	cp src/scripts/anaconda-cockpit-conf-merge $(DESTDIR)/usr/libexec/anaconda/
 	mkdir -p $(DESTDIR)/usr/lib/systemd/system/
 	cp src/systemd/webui-cockpit-ws.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth@.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth.socket $(DESTDIR)/usr/lib/systemd/system/
 	mkdir -p $(DESTDIR)/etc/anaconda/cockpit/conf.d/
 	cp src/config/cockpit/cockpit.conf $(DESTDIR)/etc/anaconda/cockpit/cockpit.conf
+	# Staged config, not active by default — webui-desktop copies it to conf.d/ at runtime
+	mkdir -p $(DESTDIR)/usr/share/anaconda/cockpit/conf.d/
+	cp src/config/cockpit/conf.d/50-remote-auth.conf $(DESTDIR)/usr/share/anaconda/cockpit/conf.d/
 
 dist: $(TARFILE)
 	@ls -1 $(TARFILE)
--- a/src/components/login/LoginPage.jsx
+++ b/src/components/login/LoginPage.jsx
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2026 Red Hat, Inc.
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+import cockpit from "cockpit";
+
+import React, { useState } from "react";
+import {
+    ActionGroup,
+    Button,
+    Form,
+    FormGroup,
+    FormHelperText,
+} from "@patternfly/react-core/dist/esm/components";
+import {
+    Card,
+    CardBody,
+    CardTitle
+} from "@patternfly/react-core/dist/esm/components/Card";
+import { HelperText, HelperTextItem } from "@patternfly/react-core/dist/esm/components/HelperText";
+import {
+    TextInputGroup,
+    TextInputGroupMain,
+    TextInputGroupUtilities
+} from "@patternfly/react-core/dist/esm/components/TextInputGroup";
+import { Bullseye } from "@patternfly/react-core/dist/esm/layouts/Bullseye";
+import { ExclamationCircleIcon } from "@patternfly/react-icons/dist/esm/icons/exclamation-circle-icon";
+import { EyeIcon } from "@patternfly/react-icons/dist/esm/icons/eye-icon";
+import { EyeSlashIcon } from "@patternfly/react-icons/dist/esm/icons/eye-slash-icon";
+
+const _ = cockpit.gettext;
+
+export const LoginPage = () => {
+    const [pin, setPin] = useState("");
+    const [error, setError] = useState("");
+    const [isLoading, setIsLoading] = useState(false);
+    const [pinHidden, setPinHidden] = useState(true);
+
+    const handleSubmit = async (event) => {
+        event.preventDefault();
+        setError("");
+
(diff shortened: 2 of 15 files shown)


Makefile
```diff
@@ -128,11 +128,21 @@ install: $(DIST_TEST) po/LINGUAS
...
+	cp src/scripts/cockpit-pin-auth $(DESTDIR)/usr/libexec/anaconda/
 	cp src/scripts/anaconda-cockpit-conf-merge $(DESTDIR)/usr/libexec/anaconda/
 	mkdir -p $(DESTDIR)/usr/lib/systemd/system/
 	cp src/systemd/webui-cockpit-ws.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth@.service $(DESTDIR)/usr/lib/systemd/system/
+	cp src/systemd/cockpit-pin-auth.socket $(DESTDIR)/usr/lib/systemd/system/
+	# static files for custom login page must be on this destination
+	mkdir -p $(DESTDIR)/usr/share/cockpit/static/
+	ln -sf ../anaconda-webui/login.js $(DESTDIR)/usr/share/cockpit/static/login-anaconda.js
+	ln -sf ../anaconda-webui/login.css $(DESTDIR)/usr/share/cockpit/static/login-anaconda.css
```
bruno-fs: @jelly is there a cleaner way of adding other assets to the custom login page?
I only managed to inject these by moving them to this folder, which seems consistent with the description of `CustomLoginPage` in the manpage:

> CustomLoginPage
>
>    Load the login page from an alternative directory. The directory must contain a login.html and optionally a po.js file for translations. When not set, the hardcoded $prefix/share/cockpit/static directory is used.

bruno-fs: solved! thanks for the help :)

src/scripts/cockpit-pin-auth
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2025 Red Hat, Inc.
```
M4rtinK: I guess we should bump this to 2026 ?

src/scripts/cockpit-pin-auth
```diff
@@ -0,0 +1,139 @@
...
+            expected = f.read().strip()
+    except FileNotFoundError:
+        send_problem_init("internal-error", f"Password file {PASSWORD_FILE} not found", {})
+        return
+
+    if password != expected:
+        send_problem_init("authentication-failed", "PIN did not match", {"password": "denied"})
+        return
+
+    os.execlpe("python3", "python3", "-m", "cockpit.bridge", os.environ)
```
M4rtinK, line 215: Why is this launching cockpit bridge ? Shouldn't that be already running by this point ? Or is it something else ?

src/scripts/cockpit-pin-auth
```diff
@@ -0,0 +1,139 @@
...
+import logging
+import os
+import re
+import sys
+import time
+
+logger = logging.getLogger(__name__)
+PASSWORD_FILE = "/run/anaconda/remote-password"
+
+
```
M4rtinK, line 80: I think this function and all the others need a comprehensive docstring, not to mention comments for all the data manipulation. Also it would be nice to have a separate description (say in the header of the file), describing how the whole thing actually works end-to-end.

The whole think seems rather complex and we really need to be able to tell what is happening (e.g. what data, in which format, moves through the system & how it is manipulated) to keep this maintainable, not to mention secure, in the long term. :)

src/scripts/cockpit-pin-auth
```diff
@@ -0,0 +1,139 @@
...
+    return response
+
+
+def decode_basic_header(response):
+    b64_encoded_header = re.match(r"^Basic\s+(.+)\s*$", response)
+    if not b64_encoded_header:
+        raise ValueError(f"Invalid Basic auth header: {response}")
+
+    decoded = base64.b64decode(b64_encoded_header.group(1)).decode()
+    # Standard Basic auth format is user:password — extract password after the colon
```
M4rtinK: If this is based on some sort of standard - please link to that standard. Also, maybe there is a library we could use for this (if its some standard authentication method) instead of all the custom code ?

src/components/login/LoginPage.jsx
```diff
@@ -0,0 +1,127 @@
...
+                                  name="pin"
+                                  value={pin}
+                                  onChange={(_event, value) => setPin(value)}
+                                  placeholder={_("Enter PIN")}
+                                />
+                                <TextInputGroupUtilities>
+                                    <Button
+                                      variant="plain"
+                                      onClick={() => setPinHidden(!pinHidden)}
+                                      aria-label={pinHidden ? _("Show password") : _("Hide password")}
```
M4rtinK: I think this might be a copy-paste error & there should "Show PIN" / "Hide PIN" ?

M4rtinK (changes requested):
While I think this is fine in general, I really thing we need to fully document the pin handling Python script. Once that is handled I think we are good to go. :)

src/scripts/cockpit-pin-auth
bruno-fs: That's a fair question, and I honestly don't know the answer. I took this script from @jelly proof of concept (that's why I added him as coauthor XD)

src/scripts/cockpit-pin-auth
bruno-fs: same for other file... funny thing I pointed the same issue with copyright headers on another PR these days 😅

src/scripts/cockpit-pin-auth
bruno-fs: I will dig here, maybe this bit is not required... if I can't figure it out will ask him

M4rtinK: Alternatively, I think we can make this a followup task (to document the script) - it is not strictly necessary for the change proposal we submitted, but we really should make sure to document the script soon. :)

src/components/login/LoginPage.jsx
bruno-fs: I actually was mixing both terms. I even called the file created by the backend PASSWORD_FILE when it should be PIN_FILE heh

M4rtinK (approved):
Looking good now, thanks a lot! :)

KKoukiou:
@bruno-fs this was merged with a test failure https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-1409-5b5662ce-20260812-152610-fedora-rawhide-boot-other/log.html which is not present in all CI runs. :/

Screenshot 2026-08-13 at 08-59-36 Require authentication for remote installation by bruno-fs · Pull Request #1409 · rhinstaller_anaconda-webui

KKoukiou:
Needs: https://github.com/rhinstaller/anaconda-webui/pull/1416 and https://github.com/rhinstaller/anaconda/pull/7228 had to be merged first.

PR #1410, opened 2026-08-11
fix: update eslint scripts to include build.js for linting

bruno-fs:
for some reason build.js wasn't included on eslint checks 😅 

EDIT: wow, everything is breaking now. maybe that's why it wasn't included. Funily enough if I run `npm run eslint` it works as expected...


bruno-fs:
make codecheck already checks build.js, so I guess we don't actually need this...

PR #1412, opened 2026-08-11
feat: use InstallationStatus as source of truth for page navigation

bruno-fs:
The remote access gives users the opportunity to access any url, which might lead to errors if anaconda is not at the proper state to show that screen. One bug resulting from this, for instance, is forcibly going to screen-progress page while the installation has not started yet: it will trigger a nasty error reported on INSTALLER-4825.

This opens an avenue of bugs that we likely need to fix anyway, but instead of patching them one by one now, this proposal is to fix them all by relying on the the installation status now reported by the backend.

**Maybe we could even rely on INSTALLATION_STATUS to reactively change to certain pages in the future instead of making individual calls to cockpit.location.go("somepage").**

Also fix a bug related to remote installation: when reloading/reconnecting after an installation already started, the Payloads module often errors with "UnknownCompsEnvironmentError". This error is 100% reproducible for dnf payload type and the installation is completed with success.

Resolves: INSTALLER-4825

### DIFF
--- a/src/apis/boss.js
+++ b/src/apis/boss.js
@@ -112,6 +112,23 @@ export const installWithTasks = () => {
 };
 
 /**
+ * @returns {Promise}           Resolves when the first installation task has been started
+ */
+export const startInstallation = () => {
+    return installWithTasks().then(tasks => {
+        const taskProxy = new BossClient().client.proxy(
+            "org.fedoraproject.Anaconda.Task",
+            tasks[0]
+        );
+        return new Promise((resolve, reject) => {
+            taskProxy.wait(() => {
+                taskProxy.Start().then(resolve, reject);
+            });
+        });
+    });
+};
+
+/**
  * @param {string} locale       Locale id
  */
 export const setLocale = ({ locale }) => {
--- a/src/apis/boss.js
+++ b/src/apis/boss.js
@@ -11,6 +11,7 @@ import { error } from "../helpers/log.js";
 import { _callClient, _getProperty } from "./helpers.js";
 
 import { moduleClients } from "./index.js";
+import { PayloadsClient } from "./payloads.js";
 
 const OBJECT_PATH = "/org/fedoraproject/Anaconda/Boss";
 const INTERFACE_NAME = "org.fedoraproject.Anaconda.Boss";
@@ -44,13 +45,21 @@ export class BossClient {
         this.dispatch = dispatch;
     }
 
-    init (args = {}) {
+    async init (args = {}) {
         this.client.addEventListener("close", () => error("Boss client closed"));
+
+        const status = await getInstallationStatus();
+        // PayloadsClient is problematic outside NOT_STARTED state, potentially leading
+        // to UnknownCompsEnvironmentError.
+        const payloadsNotNeeded = status !== INSTALLATION_STATUS.NOT_STARTED;
+        const clients = payloadsNotNeeded
+            ? moduleClients.filter(C => C !== PayloadsClient)
+            : moduleClients;
 
         return Promise.all([
             this.dispatch(getInstallationStatusAction()),
             this.dispatch(getPendingErrorAction()),
-            ...moduleClients.map(Client => new Client(this.address, this.dispatch).init(args))
+            ...clients.map(Client => new Client(this.address, this.dispatch).init(args))
         ]).then(() => {
             this.startEventMonitor();
         });
--- a/src/apis/boss.js
+++ b/src/apis/boss.js
@@ -11,7 +11,9 @@ import { error } from "../helpers/log.js";
 import { _callClient, _getProperty } from "./helpers.js";
 
 import { moduleClients } from "./index.js";
-import { PayloadsClient } from "./payloads.js";
+import { LocalizationClient } from "./localization.js";
+import { NetworkClient } from "./network.js";
+import { RuntimeClient } from "./runtime.js";
 
 const OBJECT_PATH = "/org/fedoraproject/Anaconda/Boss";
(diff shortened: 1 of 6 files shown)


src/apis/boss.js
```diff
@@ -44,13 +45,21 @@ export class BossClient {
...
+    async init (args = {}) {
         this.client.addEventListener("close", () => error("Boss client closed"));
 
+        const status = await getInstallationStatus();
+        // PayloadsClient is problematic outside NOT_STARTED state, potentially leading
+        // to UnknownCompsEnvironmentError.
+        const payloadsNotNeeded = status !== INSTALLATION_STATUS.NOT_STARTED;
+        const clients = payloadsNotNeeded
+            ? moduleClients.filter(C => C !== PayloadsClient)
+            : moduleClients;
```
bruno-fs: In a DM @KKoukiou pointed this is not an actual fix and no module should be loading after the installation has started. I will try to fix this in this PR or just drop this commit and leave it to be fixed as a follow up

src/components/app.jsx
```diff
@@ -74,6 +74,22 @@ export const Application = ({ conf, dispatch, installationStatus, isFetching, on
                 }, onCritFail({ context: N_("Reading information about the computer failed.") }));
     }, [address, automatedInstall, conf, dispatch, onCritFail]);
 
+    // Redirect to the correct page based on installation status
```
KKoukiou: I think this is not the correct approach. The user should not be able to control the installation steps from the URL in the first place because we require visiting all steps for moving forward. We should not allow the user to 'just' jump in the step because we can't know they visited everything before. 

So if the installation has started immediately jump to final step.
If the installation has not started 
* if it's KS installation to to review
* if it's non KS installation go to first screen.

The code that needs to be fixed lives in the file and is this:
```
 61     useEffect(() => {                                                                              
 62         if (path[0] && path[0] !== currentStepId) {                                                
 63             // If path is set respect it                                                           
 64             setCurrentStepId(path[0]);                                                             
 65         } else if (!currentStepId) {                                                               
 66             // Otherwise set the first step as the current step                                    
 67             setCurrentStepId(firstStepId);                                                         
 68         }                                                                                          
 69     }, [currentStepId, firstStepId, path, setCurrentStepId]);           
```

bruno-fs: I don't agree the solution has to live in (exclusively) in the wizard, as it requires a bigger context than the progress page.

When I started this implementation I had a version that was on that very same AnacondaWizard and it was quite ineffective: it would only work if the user hits "/". if user tries to directly jump to progress page, we would need to implement something similar there. 

Furthermore, and most importantly, implementing this feature on AnacondaWizard **does not fix the bug**. If someone tries to jump directly to `/anaconda-screen-progress` without a disk mounted, it will just break.

One could even argue that putting it here is hiding bugs in other places, but, architecturally speaking, I think app.jsx is the best spot to make the decision of what page to load: before any page has loaded. Don't load unnecessary context.  

So, I will change the current code to match the part of the argument I agree: don't allow users to control where they should be jumping. If it should jump to final step, this code solves it, loading the progress page. If it needs to go to wizard, then the code that lives on wizard solves the rest of it.

bruno-fs: AHA, turns out all that was missing on AnacondaWizard was another guard

https://github.com/rhinstaller/anaconda-webui/blob/7723249a440050833ac2eace16cd10d5c5e8228e/src/components/AnacondaWizard.jsx#L136

checking for the installation status here patches the bug!

bruno-fs: So I was wrong on my point about app.jsx being the best achitectural place to keep it. I thought progress page was not part of the wizard, but it also is. So that's the correct place to keep the logic. all it was missing was the guard, and I also wasn't aware about this path that loads the page with less context

https://github.com/rhinstaller/anaconda-webui/blob/7723249a440050833ac2eace16cd10d5c5e8228e/src/components/AnacondaWizard.jsx#L137-L140

bruno-fs: so I went ahead and tried the reactive approach you suggested - moving the logic to AnacondaWizard.

  it works. the design is cleaner. but it was way bigger than expected. here's what was needed:

  the core idea: instead of navigating to the progress page directly, the review page calls startInstallation() which starts the
  backend process. the wizard then reacts to installationStatus changing and navigates to the progress page only when the
  status is no longer NOT_STARTED. this way a synchronous guard works - by the time the progress page renders, the status has
  already updated.

  what I had to change:
  - AnacondaWizard.jsx - unified navigation useEffect that reacts to installationStatus, plus a guard that prevents rendering
  the progress page when status is NOT_STARTED
  - ReviewConfiguration.jsx - both the auto-proceed (L134) and the button click (L358) now call beginInstallation() instead of
  navigating directly
  - InstallationProgress.jsx - removed startNewInstallation() from the default case. with the wizard guard, reaching the
  progress page with NOT_STARTED is now impossible, so the default case is now an error
  - boss.js - added startInstallation() function, added InstallationStatus handling to the event monitor

  what I didn't expect: PropertiesChanged signals were never being emitted for Boss properties. watch_property in dasbus only
  caches changes - it needs flush_changes() to actually emit the DBus signal. the other modules (localization, network, etc.)
  get this from ModuleInterfaceTemplate, but BossInterface inherits plain InterfaceTemplate which doesn't wire it up. so this
  also needs a backend fix: connecting module_properties_changed to flush_changes in BossInterface, and emitting
  module_properties_changed after property changes in Boss.

  tl;dr: the approach works and is better, but it turned a 2-point fix into a cross-repo change with a backend fix.

KKoukiou:
Closing in favor of: https://github.com/rhinstaller/anaconda-webui/pull/1422

PR #1413, opened 2026-08-13
Add TLS proxy for remote installation

bruno-fs:
Add a lightweight Python asyncio TLS proxy that terminates TLS on
port 443 and forwards plaintext TCP to cockpit-ws on localhost:80.
This enables HTTPS for remote installation without requiring the
full cockpit-tls stack.

Reuses cockpit infrastructure where possible:
- cockpit-certificate-ensure generates the self-signed certificate
- --for-tls-proxy on cockpit-ws relaxes Origin checking for WebSocket
  connections behind the proxy

When remote PIN is set, cockpit-ws binds to 127.0.0.1:80 (not externally
accessible) and the TLS proxy handles external connections on port 443.
The plain HTTP path on port 80 remains unchanged for local installation
or remote access without a PIN set.

Assisted-by: Claude Opus 4.6
Resolves: INSTALLER-4617


src/systemd/webui-tls-proxy.service
```diff
@@ -0,0 +1,16 @@
...
+Requires=webui-cockpit-ws.service
+
+[Service]
+# RuntimeDirectory is required by cockpit-certificate-ensure
+RuntimeDirectory=cockpit/tls
+# Cert is saved to /etc/cockpit/ws-certs.d/ (hardcoded in cockpit)
+ExecStartPre=+/usr/libexec/cockpit-certificate-ensure --for-cockpit-tls
+ExecStart=/usr/bin/python3 /usr/libexec/anaconda/webui-tls-proxy
+Restart=on-failure
+RestartSec=1
```
jelly, line 13: cockpit-tls is quite hardened, this runs as root so you probably could/want to apply some [systemd hardening](https://wiki.archlinux.org/title/Systemd/Sandboxing)

M4rtinK (approved):

test/check-remote-auth
```diff
@@ -19,34 +18,41 @@ class TestRemoteAuth(VirtInstallMachineCase):
...
+        # Opening at "/" should work (using webui_path should work too, but that's given)
+        b.open("/", tls=True)
 
         # Login page should be visible
         b.wait_visible("#pin-input")
 
         # Wrong password should be rejected
         b.set_input_text("#pin-input", "wrongpin")
         b.click("button[type=submit]")
-        b.wait_in_text(".pf-m-error", "Invalid PIN")
```
KKoukiou, line 42: This change is wrong.

PR #1416, opened 2026-08-13
tests: fix selector for helper text in pin auth screen

KKoukiou:


M4rtinK (approved):
Looking good, thanks! :)

PR #1417, opened 2026-08-13
localization,datetime: mark incomplete kickstart pages as needing user visit

KKoukiou:


adamkankovsky (changes requested):
The Kickstart files are now out of date following this change. The tests are failing.

adamkankovsky (approved):
LGTM

PR #1420, opened 2026-08-17
spec: require slitherer for select spins, firefox for the rest

KKoukiou:
Define webui_slitherer_spin_identities macro for spins that should use slitherer instead of firefox (budgie, cinnamon, kde-desktop, kde-mobile, lxde, lxqt, matecompiz, sway, xfce, i3).


M4rtinK (approved):
Looking good as far as I can tell, thanks. :)

PR #1428, opened 2026-08-20
tests: ignore the header text in the pixel tests

KKoukiou:
This is the os-release pretty name which get's bumped every release. Let's reduce the noise.

See https://github.com/cockpit-project/bots/pull/9356


rvykydal (approved):
Looks good to me.

KKoukiou:
Mergin even with the one test failure because this enables the image refresh and it's critical for resolving CI instability at the moment.

PR #1429, opened 2026-08-21
tests: when reading default timezone split only on first occurance of / delimiter

KKoukiou:
Some timezones specify City/Region. Example: America/Argentina/Buenos Aires.


rvykydal (approved):
Looks good to me

PR #1431, opened 2026-08-24
storage: only fetch free/total space for disk devices (gh1715)

adamkankovsky:
GetDiskFreeSpace and GetDiskTotalSpace are only valid for disk devices. Calling them for LUKS or other non-disk devices causes an UnknownDeviceError traceback in anaconda, which kstest detects as a test failure.

Guard the calls with devData["is-disk"].v to skip non-disk devices.

### DIFF
--- a/src/actions/storage-actions.js
+++ b/src/actions/storage-actions.js
@@ -42,12 +42,17 @@ export const getDevicesAction = () => {
             try {
                 const devData = await getDeviceData({ disk: device });
 
-                const free = await getDiskFreeSpace({ diskNames: [device] });
-                // extend it with variants to keep the format consistent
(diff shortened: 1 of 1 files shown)


KKoukiou (changes requested):
Try setting the values to 0 in the case of no disk https://github.com/rhinstaller/anaconda-webui/pull/1383/changes/f3520a9d86e07d1c5a6975da4482d310e033e39b

KKoukiou:
> Try setting the values to 0 in the case of no disk [f3520a9](https://github.com/rhinstaller/anaconda-webui/commit/f3520a9d86e07d1c5a6975da4482d310e033e39b)

Sorry wrong suggestion

KKoukiou (approved):

PR #1432, opened 2026-08-24
Fix kickstart installations without storage and disable forward sidebar navigation

KKoukiou:


src/components/storage/installation-method/usePageComplete.jsx
```diff
@@ -47,10 +47,8 @@ export const useStorageComplete = () => {
         requiredSpace != null && freeSpace != null && requiredSpace > freeSpace;
 
     const pending =
-        !appliedPartitioning ||
```
rvykydal, line 50: I am wondering if by removing this guard we can't break some non-kickstart use case (like the state being validly pending)?

KKoukiou: Fixed. Checking only if it's undefined (not set yet by API) and if set to '' we can continue.

adamkankovsky (approved):
If those tests are successful, that seems fine to me

KKoukiou:
Unrelated test failure

PR #1436, opened 2026-08-25
fix: only enable PIN auth for remote installations

KKoukiou:
WEBUI_NOAUTH defaults to 0, so the previous condition enabled the remote auth config for all installations, including local ones. Gate on WEBUI_REMOTE=1 as well so local installations skip the login page.


rvykydal (approved):

KKoukiou:
Unrelated test failure.

PR #1446, opened 2026-08-31
fix: preserve installation progress step across page reloads

tomasfratrik:
The progress stepper reset to "Storage configuration" on page reload or when opening the installer in a different browser, because the current category is only pushed through D-Bus signals and can't be queried after the fact.

Persist the current progress step to /run/anaconda/webui-progress-step via cockpit.file() so any connecting browser can restore it.

Assisted-by: Claude Opus 4.6
Resolves: INSTALLER-4872

### DIFF
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -94,15 +94,17 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                         context: cockpit.format(N_("Installation of the system failed: $0"), refStatusMessage.current),
                     }));
                 });
-                categoryProxy.addEventListener("CategoryChanged", (_, category) => {
-                    debug("CategoryChanged:", category);
-                    const step = progressStepsMap[category];
-                    setCurrentProgressStep(current => {
-                        if (step !== undefined && step >= current) {
-                            return step;
-                        }
-                        return current;
-                    });
+                categoryProxy.addEventListener("changed", (_, data) => {
+                    if ("CurrentCategory" in data) {
+                        debug("CategoryChanged:", data.CurrentCategory);
+                        const step = progressStepsMap[data.CurrentCategory];
+                        setCurrentProgressStep(current => {
+                            if (step !== undefined && step >= current) {
+                                return step;
+                            }
+                            return current;
+                        });
+                    }
                 });
                 categoryProxy.addEventListener("ErrorRaised", (_, message, detailType) => {
                     handleError(message, detailType, categoryProxy);
@@ -113,11 +115,15 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                     setCurrentProgressStep(PROGRESS_STEPS_DONE);
                 });
             };
-            taskProxy.wait(() => {
+            Promise.all([taskProxy.wait(), categoryProxy.wait()]).then(() => {
                 addEventListeners();
                 if (shouldStart) {
                     taskProxy.Start().catch(onCritFail(failureCtx));
                 } else {
+                    const step = progressStepsMap[categoryProxy.CurrentCategory];
+                    if (step !== undefined) {
+                        setCurrentProgressStep(step);
+                    }
                     getSteps({ task: taskPath })
                             .then(
                                 ret => setSteps(ret.v),


tomasfratrik:
Also this logic can be done throughout backend. Note that in webui you can't use session/local storage because that obviously doesn't persist when chaning another browser, so this is maybe the only solution for webui.

tomasfratrik:
Here is example of such backend implementation: https://github.com/rhinstaller/anaconda/pull/7281 which exposes current stage via the dbus interface

tomasfratrik:
I will check the bot's comments later after we decide if the solution can be here or it should be in backend

tomasfratrik:
changing this PR to connect to the backend solution https://github.com/rhinstaller/anaconda/pull/7281

src/components/installation/InstallationProgress.jsx
```diff
@@ -113,11 +115,15 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                     setCurrentProgressStep(PROGRESS_STEPS_DONE);
                 });
             };
-            taskProxy.wait(() => {
+            Promise.all([taskProxy.wait(), categoryProxy.wait()]).then(() => {
                 addEventListeners();
                 if (shouldStart) {
```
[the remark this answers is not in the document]
tomasfratrik: Okay If I get it this race condition can happen (?), so I will impelemnt the suggested fix

src/components/installation/InstallationProgress.jsx
```diff
@@ -113,11 +115,15 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                     setCurrentProgressStep(PROGRESS_STEPS_DONE);
                 });
             };
-            taskProxy.wait(() => {
-                addEventListeners();
+            addEventListeners();
+            Promise.all([taskProxy.wait(), categoryProxy.wait()]).then(() => {
                 if (shouldStart) {
```
bruno-fs:
```suggestion
            Promise.all([taskProxy.wait(), categoryProxy.wait()]).then(() => {
                addEventListeners();
                if (shouldStart) {
```
shouldn't we run addEventListeners only after taskProxy.wait (and now categoryProxy.wait) resolves? I honestly don't understand the implications of this change, but the original implementation was doing it like that, which makes me think it was intentional to keep it that way

bruno-fs (commented):
I'm fully onboard with this approach and the required changes on backend! 🚀 

I think these changes paired with properties for error messages are an amazing tool that would allow for a future refactor on progress page, deduplicating a lot of code.

We could change the reliance on signals to instead just follow the properties. I wouldn't change that on the context of this fix, because I think it would be a big refactor in order to keep the useEffect side effects minimal. WDYT, @KKoukiou?

src/components/installation/InstallationProgress.jsx
tomasfratrik: This was a bot proposed change, let me analyze it again, this was something about race condition I think, I need to get back to this and analyze it

bruno-fs: we have a tendency to ignore bot suggestions if it doesn't provide a compelling reason for the change

tomasfratrik: both waits should start at the same time as as I get it, so the gap is in miliseconds. I will revert the change as you suggested

bruno-fs (approved):
Looks good! thanks :)

bruno-fs:
Including the following patch in this proposal would help preventing future regressions

```patch
diff --git a/src/components/installation/InstallationProgress.jsx b/src/components/installation/InstallationProgress.jsx
index 2d53885c..2a02b385 100644
--- a/src/components/installation/InstallationProgress.jsx
+++ b/src/components/installation/InstallationProgress.jsx
@@ -297,7 +297,7 @@ export const InstallationProgress = ({ automatedInstall, onCritFail }) => {
                                             description={
                                                 <Flex direction={{ default: "column" }}>
                                                     <FlexItem spacer={{ default: "spacerNone" }}>
-                                                        <Content component="p">{phaseText}</Content>
+                                                        <Content ouiaId={progressStep.id + "-phase"} component="p">{phaseText}</Content>
                                                     </FlexItem>
                                                     <FlexItem spacer={{ default: "spacerNone" }}>
                                                         <Content component="p">{statusText}</Content>
diff --git a/test/check-progress b/test/check-progress
index c1d1cfe3..0e08eaa4 100755
--- a/test/check-progress
+++ b/test/check-progress
@@ -40,10 +40,17 @@ class TestInstallationProgress(VirtInstallMachineCase):
         with b.wait_timeout(300):
             b.wait_in_text("h2", "Installing")
 
+        # Wait until payload step transition to "in progress" to verify reconnect. This particular step
+        # was chosen because it should take longer than others to run - making it less prone to race
+        # when reloading
+        payload_status_selector = "p[data-ouia-component-id='anaconda-screen-progress-step-payload-phase']"
+        with b.wait_timeout(100):
+            b.wait_in_text(payload_status_selector, "In progress")
+
         # Reload the page while installation is running to verify reconnect
         b.reload()
         b.wait_visible(".pf-v6-c-progress-stepper")
-        b.wait_in_text("h2", "Installing")
+        b.wait_in_text(payload_status_selector, "In progress")
 
         p.wait_done()
         b.wait_in_text("h2", "Successfully installed")

```

bruno-fs:
Looks good, but I think there's room for improvement here:

1. please add the suggested changes to the test case so we can be (somewhat) protected from future regressions (see [this](https://github.com/rhinstaller/anaconda-webui/pull/1446#issuecomment-5552086196))
2.  add some explanation to the body of the new commit "Introduce new install step". Maybe something like "Final installation phase was misconfigured and missing a step, causing the stepper to go back a phase on reconnections".

KKoukiou (approved):

tomasfratrik:
> Looks good, but I think there's room for improvement here:
> 
>     1. please add the suggested changes to the test case so we can be (somewhat) protected from future regressions (see [this](https://github.com/rhinstaller/anaconda-webui/pull/1446#issuecomment-5552086196))
> 
>     2. add some explanation to the body of the new commit "Introduce new install step". Maybe something like "Final installation phase was misconfigured and missing a step, causing the stepper to go back a phase on reconnections".

Yes your proposed change for test, looks good, so I implemented this, and also fixed the commit message.

tomasfratrik:
for explanation why steps changed check comments in the PR that is connected with this one: https://github.com/rhinstaller/anaconda/pull/7281

tomasfratrik:
the test fail I think is because the backend is not merged

tomasfratrik:
Failing test doesn't have backend, the test with it `fedora-rawhide-boot/other-anaconda-pr-7281` passes

PR #1447, opened 2026-08-31
fix stuck progress screen on ostree installs

adamkankovsky:


KKoukiou (commented):
Please put a Resolves: rhbz line.

KKoukiou (commented):
This might fix the issue but it seems like paper over the real problem. 

I would suggest to switch your fix to change startNewInstallation method to only starts the installation task. Listeners should be attached once, when installationStatus becomes RUNNING.

AdamWill:
Bunch of tests running at https://openqa.stg.fedoraproject.org/tests/overview?distri=fedora&version=46&build=Kojitask-149769836-NOREPORT&groupid=2 , if they all pass it means this looks promising. If so I'll re-run the set a few more times to make sure.

KKoukiou (commented):
I don't think that's a valid situation described in the commit message. We can't just 'lose' events.
Under heavy load maybe things are slower, not lost.

AdamWill:
The version of this I tested last night didn't fix it. Out of 12 runs, 2 got stuck: https://openqa.stg.fedoraproject.org/tests/6914506 and https://openqa.stg.fedoraproject.org/tests/6914516 .

I see it's grown an extra commit today. Do you want me to try this version, or does @KKoukiou 's comment mean this is not worth testing?

KKoukiou:
@AdamWill we would appreciate to  see the extra logs to debug this further which were added in webui latest update: https://bodhi.fedoraproject.org/updates/FEDORA-2026-cb31689746

KKoukiou:
> @AdamWill we would appreciate to see the extra logs to debug this further which were added in webui latest update: https://bodhi.fedoraproject.org/updates/FEDORA-2026-cb31689746

From that update https://openqa.fedoraproject.org/tests/5204051 this test in question succeeded. Can I keep retriggering till I get a failed run?

AdamWill:
Just for the record, per Matrix discussion, logs from a failure with anaconda-webui-82-1.fc46 are available [here](https://openqa.fedoraproject.org/tests/5237765#downloads).

AdamWill:
I did also test the updated version of the PR, and it seems to fail in an [interestingly different way](https://openqa.stg.fedoraproject.org/tests/6917032#step/_do_install_and_reboot/12) - instead of hanging, we now get a "Connection has timed out." error. It still doesn't fail on every try, some attempts succeeded.

KKoukiou:
This issue on Silverblue installation should be fixed with https://github.com/rhinstaller/anaconda-webui/pull/1457

AdamWill:
Great, thanks. I'll yell if I still see the problem happening. Can we have that fix for F45 also? Do I need to file an FE?

PR #1449, opened 2026-08-31
Add option to attach disk to testvm

tomasfratrik:
Pass --disk args to virt-install so disks are part of the domain definition from the start. Anaconda discovers them during its initial storage scan. Extra disks are created as temporary qcow2 images and cleaned up when the VM is destroyed.

This feature can be used with --add-disk <size in GiB> while using webui_testvm.py

Example:
`./test/webui_testvm.py fedora-rawhide-boot --add-disk 15`

This change is useful for testing purposes

This is not part of any ticket

### DIFF
--- a/test/machine_install.py
+++ b/test/machine_install.py
@@ -8,6 +8,7 @@ import shlex
 import socket
 import subprocess
 import sys
+import tempfile
 import time
 from tempfile import TemporaryDirectory
 
@@ -46,6 +47,7 @@ class VirtInstallMachine(VirtMachine):
         self.pause_at_summary = kwargs.pop("pause_at_summary", False)
         self.payload_type = kwargs.pop("payload_type", "liveimg".lower())
         self.remote_pin = kwargs.pop("remote_pin", "")
+        self.extra_disks = kwargs.pop("extra_disks", [])
         # Always enforce the minimum RAM the installer requires. Covers every entry point the same way:
         # test/run (CI), local test/check-*, and GlobalMachine.reset() which omits memory_mb.
         kwargs["memory_mb"] = max(kwargs.get("memory_mb") or 0, INSTALLER_VM_MEMORY_MB)
@@ -196,7 +198,18 @@ class VirtInstallMachine(VirtMachine):
             else ""
         )
         auth_method = f"inst.webui.remote.pin={self.remote_pin}" if self.remote_pin else "inst.webui.remote.noauth"
+
         try:
+            disk_args = "--disk=none "
+            if self.extra_disks:
+                disk_parts = []
+                for size_gb in self.extra_disks:
+                    fd, path = tempfile.mkstemp(suffix='.qcow2', prefix=f"disk-anaconda-{self.label}-")
+                    os.close(fd)
+                    os.unlink(path)
+                    disk_parts.append(f"--disk path={path},size={size_gb},bus=virtio ")
+                disk_args = "".join(disk_parts)
+
             self._execute(
                 "virt-install "
                 "--wait "
@@ -221,7 +234,7 @@ class VirtInstallMachine(VirtMachine):
                 "-device virtio-net-pci,netdev=hostnet0,id=net0,addr=0x16' "
                 f"--extra-args '{extra_args}' "
                 f"{extra_boot_args_option}"
-                f"--disk=none "
+                f"{disk_args}"
                 f"--location {location} &"
             )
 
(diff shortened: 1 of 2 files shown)


test/machine_install.py
```diff
@@ -196,6 +198,20 @@ def start(self):
...
+
+        self._disk_images = []
+        disk_args = "--disk=none "
+        if self.extra_disks:
+            disk_parts = []
+            for index, size_gb in enumerate(self.extra_disks):
+                fd, path = tempfile.mkstemp(suffix='.qcow2', prefix=f"disk-anaconda-{self.label}-")
+                os.close(fd)
+                subprocess.check_call(["qemu-img", "create", "-f", "qcow2", path, f"{size_gb}G"])
+                self._disk_images.append(path)
```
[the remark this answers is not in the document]
tomasfratrik: I think valid. will fix

test/webui_testvm.py
```diff
@@ -23,13 +23,17 @@ def cmd_cli():
                         action='store_true', dest="pause_at_summary")
     parser.add_argument("--remote-pin", help="PIN for accessing the webui. Blank value defaults to inst.webui.remote.noauth",
                         dest="remote_pin")
+    parser.add_argument("--add-disk", help="Attach a virtual disk with the given size in GiB (e.g., --add-disk 15)",
+                        type=int, metavar="SIZE", dest="add_disk")
     args = parser.parse_args()
 
     if args.bios:
         os.environ["TEST_FIRMWARE"] = "bios"
+    extra_disks = [args.add_disk] if args.add_disk else []
```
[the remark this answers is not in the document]
tomasfratrik: will fix, someone can pass 0

test/machine_install.py
```diff
@@ -196,7 +198,22 @@ def start(self):
...
+        self._disk_images = []
+
         try:
+            disk_args = "--disk=none "
+            if self.extra_disks:
+                disk_parts = []
+                for index, size_gb in enumerate(self.extra_disks):
+                    fd, path = tempfile.mkstemp(suffix='.qcow2', prefix=f"disk-anaconda-{self.label}-")
+                    os.close(fd)
+                    subprocess.check_call(["qemu-img", "create", "-f", "qcow2", path, f"{size_gb}G"])
```
KKoukiou: There is no need to generated it - just pass the size and the path (default location would also work) and virt-install will do it for you. No need to specify target either.

bruno-fs: I personally like the idea of doing it manually like this, as doing it with tempfile would get rid of the disk for sure on startups. If we let virt-install manage it the disk might linger depending on how the VM is teared down.

The argument could be made that just passing the option to virt-install is good enough and less code is usually better - less maintenance. but it isn't like we are adding a lot more by doing it like this, so I'd be fine with either approach.

bruno-fs (approved):
That's certainly a very welcome QoL change =D

test/machine_install.py
KKoukiou: I don't think I follow. You can still manually cleanup disk files - the location that virt-install stores the disk files is known and static.

tomasfratrik: yes it makes more sense to just let virt-install handle it

tomasfratrik:
So I made the virt-install handle the creation, but still im thinking of implementing a way to make the qcow2 to be deleted when an error is encountered or double sigint with Ctrl+cc. Even when its in the /tmp, if you encounter many errors, it can get quite huge before some restart

tomasfratrik:
Okay I have implemented this sigint catch as another commit, tested it and the disk is removed

test/machine_install.py
bruno-fs: @KKoukiou 
> I don't think I follow. You can still manually cleanup disk files - the location that virt-install stores the disk files is known and static.

yes, you can. but do you? I only bother doing that if I look at the folder and see a bunch of unused disks. writing to tmp would be a way to get that for "free" (free = on a restart). 

current approach already handles that, so I'm fine with it too :)

test/machine_install.py
```diff
@@ -244,6 +260,10 @@ def start(self):
                 # Symlink /usr/share/cockpit to /usr/local/share/cockpit so that rsync works without killing cockpit-bridge
                 Machine.execute(self, "mkdir -p /usr/local/share/cockpit/anaconda-webui && mount --bind /usr/share/cockpit /usr/local/share/cockpit")
         except Exception as e:
+            for path in self._disk_images:
```
KKoukiou: This should be handled by --remove-all-storage bellow.

test/machine_install.py
```diff
@@ -256,6 +276,9 @@ def _cleanup(self, quick=False):
             f"virsh -q -c qemu:///session undefine --nvram "
             f"--remove-all-storage {self.label} || true"
         )
+        for path in getattr(self, "_disk_images", []):
```
KKoukiou: Same here; --remove-all-storage

test/machine_install.py
tomasfratrik: completely forgot about that, definitely better letting it hadle the removal

KKoukiou (approved):

PR #1450, opened 2026-08-31
fix: reset boot drive on disk selection change and storage rescan

KKoukiou:
Reset Drive to "" (auto-detect) when the user changes disk selection or rescans storage, matching the GTK UI's apply_disk_selection pattern.

A stale Bootloader.Drive D-Bus property from a previous configure task can cause the bootloader to target the wrong disk.


KKoukiou:
This is expected to fix this test failure / flakiness https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-1440-f7db1654-20260831-110901-fedora-rawhide-boot-storage/log.html

adamkankovsky (approved):
Thanks for fix

PR #1454, opened 2026-09-01
fix: return after onFail for bootloader on MDRAID detection

KKoukiou:
Without the return, handleMDRAID continued to setNewSelectedDisks() which called setNextCheckStep(), overriding the failed state and letting the pipeline proceed to backend validation. The backend error then replaced the MDRAID-specific error message.


KKoukiou:
See failure that it fixes here: https://github.com/rhinstaller/anaconda-webui/pull/1438

adamkankovsky (approved):
Looks great to me. Thanks

PR #1457, opened 2026-09-03
fix: disable cockpit transport health check and skip redundant re-renders

KKoukiou:


adamkankovsky (approved):
LGTM, thanks for this change

KKoukiou:
I still get health check failures 
Screenshot 2026-09-03 at 16-57-30 Virtual machines - kkoukiou@faros

KKoukiou:
Now its ignored, we can see in the logs as well:

Screenshot 2026-09-04 at 12-44-16 Virtual machines - kkoukiou@kkoukiou-thinkpadp1gen8 rmtde csb

adamkankovsky:
If that doesn't work, I'm also working on a backend change that will create a keepalive thread for the ostree installation.

KKoukiou:
Ok it works not - installation finished.

PR #1461, opened 2026-09-09
Move error to the footer so that it's sticky when dialog get's too long

KKoukiou:
Resolves: rhbz#2526398


tomasfratrik:
This fix puts the error (traceback which is also not so pretty) at the end of the screen, my question is if this expired key shouln't be handled the same way the revoked key is (it doesn't go beyond the first screen where you input the key), which in my opition is the more clean and optimal version: 
Screenshot From 2026-09-09 15-40-17

But this change is still good regardles, because any other unexpected error can be now better visible/findable

KKoukiou:
> 2526398

We don't handle this in a different way. If bugzilla login fails we show this screen you pasted. For some reason on the example on the bugzilla the login was possible and only reporting failed.

tomasfratrik:
hm that is interesting that the login was possible I cannot try this without expired key, but maybe this is intended ? like that the revoked key should represent some "non existent" key, and on the other hand expired lets you login just to tell you it is expired.. don't know just speculating.... but for sure this change is still really good for visibility of such failures (haven't tested this yet)

src/components/Error.jsx
```diff
@@ -657,6 +637,12 @@ const BZExceptionReportFlow = ({
                 </Form>
             </ModalBody>
             <ModalFooter>
+                {bugCreationError && (
+                    <Alert
+                      title={bugCreationError}
```
tomasmatus: Not sure what the `bugCreationError` can contain but I like the previous approach better. To have a useful user-facing title "Failed to report issue" and then drop the whole error in the content bellow the title :)

KKoukiou: Good point.

tomasfratrik (approved):
lgtm

PR #1464, opened 2026-09-09
test: Add Stratis test cases to check-storage-cockpit

vojtechtrefny:
Cover both encrypted and non-encrypted Stratis pools and filesystems created in Cockpit storage and consumed by Anaconda via the 'use-configured-storage' scenario.

PR #1472, opened 2026-09-14
test: add Stratis partitioning coverage via Cockpit Storage and kickstart

adamkankovsky:
Storage Editor path: add a nondestructive test that drives the Cockpit Storage editor to create a Stratis pool and root filesystem, and verifies the installer accepts the Stratis layout through to the Review screen (use-configured-storage scenario).

Kickstart path: add TestKickstartAutomatedStratis with a kickstart that uses Anaconda's native 'stratispool'/'stratisfs' commands to put /home on a Stratis pool while / stays on ext4. The test performs a full install, reboots, and verifies / is on a regular partition and /home is on the Stratis pool. Verified passing locally (fedora-45-boot).

Both test-day case definitions document a known Fedora 45 limitation: placing / directly on Stratis installs but does not boot (early boot hangs on dev-stratis-<pool>-root.device, as the pool is not activated in the initramfs), so Stratis is used only for /home in the bootable kickstart path.

Assisted-by: claude-sonnet-4-6


KKoukiou (commented):
There is already a PR from Vojtech for covering Stratis: https://github.com/rhinstaller/anaconda-webui/pull/1464

KKoukiou (changes requested):
kickstart stratis tests will be better covered in the kickstart tests repository. We can't duplicate everything here. 

Also we don't do these test case definitions in yml format, let's not introduce something without proper justification.

adamkankovsky:
Closing as redundant/superseded, based on the review:

- Stratis coverage via the Cockpit Storage editor is already provided by @vojtechtrefny's #1464 (`testStratis` and `testStratisEncrypted` in `check-storage-cockpit`), so this PR duplicates it.
- Kickstart Stratis tests are better covered in the kickstart tests repository; we should not duplicate them here.
- We don't define test cases in YAML, so the introduced `.tc.yaml` files are dropped.

Thanks @KKoukiou for the review.

PR #1474, opened 2026-09-15
storage: stop device-tree walk from racing with installation teardown

adamkankovsky:
getDevicesAction() reactively re-walks the whole storage device tree on every AppliedPartitioning signal via a long chain of sequential DBus round-trips. If installation has already started, CreateStorageLayoutTask can tear down transient devices (e.g. a LUKS container) while this walk is still in flight, since stopping the DBus subscription in boss.js does not cancel an already-running async walk. Track installation activity in a shared flag and check it both before starting a new walk and on every iteration of the in-flight one, so it bails out instead of racing the teardown - matching GTK, which never recomputes storage during install.


KKoukiou:
Let's convert to DRAFT till we have KS tests passing.

KKoukiou (approved):
A bit hackish but it looks safe enough and it shows that it resolves the UnknownDevicesError s

